PR #39, opened 2015-10-16
Add ZipCodeRow cell type

davidrothera:
As created for XLForm in https://github.com/xmartlabs/XLForm/commit/23b93a1a70bb17b8f2f65a4164978b6ac1190d9e this adds the ZipCode row type.

Example also added :+1:

### DIFF
--- a/Example/Example/ViewController.swift
+++ b/Example/Example/ViewController.swift
@@ -237,6 +237,11 @@ class RowsExampleViewController: FormViewController {
                         $0.title = "AccountRow"
                         $0.placeholder = "Placeholder"
                     }
+        
+                <<< ZipCodeRow() {
+                        $0.title = "ZipCodeRow"
+                        $0.placeholder = "90210"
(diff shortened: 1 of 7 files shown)


davidrothera:
Also, as a background to why this has been added, you need to change the keyboard and auto-capitilization properties of a textField however these settings are not observed in KVO :(

mtnbarreto:
Hey @davidrothera, Thanks for contributing! 

I think that `ASCIICapable` keyboard type makes more sense than `.Default` for a zip code row because the `.Default` type allows to enter emoji icons. What do you think?

smaspe:
@mtnbarreto How about `NamePhonePad` ? letters and numbers (bigger numbers). Punctuation is not use in zipcodes anywhere, afaik

davidrothera:
The `.NamePhonePad` option doesn't support auto-capitilization so not really ideal here

PR #58, opened 2015-10-27
First stab at a SliderRow.

SpacyRicochet:
Very much a work in progress, but I wanted to start at this if we ever update our project that uses XLForm to use Eureka instead :)

I'm open to any suggestions for improvement and I'm wondering whether this is something you guys want to see in Eureka (since it was part of XLForm), or if you want these things set up as custom cells instead.


mats-claassen:
Updated your pull request at #274

SpacyRicochet:
I see it made it in :) Glad to be of help.

PR #73, opened 2015-11-04
Adds @noescape attribute to row initializer closure

jpvasquez:
This has various advantages, see http://nshint.io/blog/2015/10/23/noescape-attribute/ and http://stackoverflow.com/questions/28427436/noescape-attribute-in-swift-1-2


mtnbarreto:
@jpvasquez thanks for contributing!

PR #77, opened 2015-11-08
Add setImage function by NSURL

ykeisuke:
I add `setImage` function.
This function is seting image by NSURL.
And can set initial Image. :octocat:


mtnbarreto:
@ykeisuke Thanks for contributing! This feature seems to me something that should no be provided by default since we can easily implement it through swift extensions and we do not always want to setup a image row using NSURL.

PR #93, opened 2015-11-18
Fix constraints update on _FieldCell and SegmentedCell

dernster:
Modified observeValueForKeyPath function on _FieldCell and SegmentedCell to correctly update constraints


mtnbarreto:
:+1:

Issue #103, opened 2015-11-26
"Clear" button on accessory navigation toolbar?


blue-toe:
Is it possible to add another button to the accessory navigation toolbar to clear the currently selected field? Right now there does not appear to be a way for the user to clear a date field (ie. I have an optional date field).

mtnbarreto:
@blue-toe 
Accessory navigation toolbar can be changed  by extending it and setting up navigationAccessoryView form controller property with your custom navigation (`MyNavigationAccessoryView`) as the code below shows.

```swift
class MyNavigationAccessoryView : NavigationAccessoryView {
    
    override init(frame: CGRect) {
        super.init(frame: CGRectMake(0, 0, frame.size.width, 44.0))
        autoresizingMask = .FlexibleWidth
        fixedSpace.width = 22.0
        setItems([previousButton, fixedSpace, nextButton, flexibleSpace, clearButton, fixedSpace, doneButton], animated: false)
    }
    
    var clearButton = UIBarButtonItem(title: "Clear", style: UIBarButtonItemStyle.Plain, target: nil, action: nil)
    private var fixedSpace = UIBarButtonItem(barButtonSystemItem: .FixedSpace, target: nil, action: nil)
    private var flexibleSpace = UIBarButtonItem(barButtonSystemItem: .FlexibleSpace, target: nil, action: nil)

    required init?(coder aDecoder: NSCoder) {
        fatalError("init(coder:) has not been implemented")
    }

}

class RowsExampleViewController: FormViewController {
    
    
    func clearAction(button: UIBarButtonItem){
        guard let currentRow = tableView?.findFirstResponder()?.formCell()?.baseRow else { return }
        currentRow.baseValue = nil
        currentRow.updateCell()
    }
    
    
    override func viewDidLoad() {
        super.viewDidLoad()
        
        
        navigationAccessoryView = {
            let naview = MyNavigationAccessoryView(frame: CGRectMake(0, 0, self.view.frame.width, 44.0))
            naview.doneButton.target = self
            naview.doneButton.action = "navigationDone:"
            naview.previousButton.target = self
            naview.previousButton.action = "navigationAction:"
            naview.nextButton.target = self
            naview.nextButton.action = "navigationAction:"
            naview.tintColor = self.view.tintColor
            naview.clearButton.target = self
            naview.clearButton.action = "clearAction:"
            return naview
        }()
       ...
       ..
       .
```

Hope it helps you.

blue-toe:
[...]

PR #117, opened 2015-12-01
Clear Button

blue-toe:
Added “Clear” button to NavigationAccessoryView. When pressed, will set
current row value to nil.


mtnbarreto:
NavigationAccessoryView is fully customizable and i think the actions (Done, previous, Next) provided by default are fine.

In the following issue conversation thread we discussed how to customize it:  https://github.com/xmartlabs/Eureka/issues/103

Regards

PR #130, opened 2015-12-09
Delete unnecessary semicolons

yuuki1224:
I think these semicolons are unnecessary unless adding additional statement after it.


mtnbarreto:
Thanks @yuuki1224 !

PR #132, opened 2015-12-09
Add missing access control

yuuki1224:
By default, access control level is `internal`, right? I've wondered only in this method, access control isn't defined explicitly. So I've added it.

### DIFF
--- a/Source/Core.swift
+++ b/Source/Core.swift
@@ -397,7 +397,7 @@ extension Form {
         }
     }
     
-    func dictionaryValuesToEvaluatePredicate() -> [String: AnyObject] {
+    internal func dictionaryValuesToEvaluatePredicate() -> [String: AnyObject] {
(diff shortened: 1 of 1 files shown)


mtnbarreto:
@yuuki1224  I there any reason you want to make the accessor level explicit? 
I would prefer to omit `internal`. I think we have been doing it throughout the base code.

yuuki1224:
Don't have particular reason. but I've just thought on the rest of parts which are used `internal` keyword(like here([1](https://github.com/xmartlabs/Eureka/blob/master/Source/Core.swift#L816), [2](https://github.com/xmartlabs/Eureka/blob/master/Source/Core.swift#L822))), they are explicitly indicated. But only this part, `internal` keyword is missing. (If you're planning to delete all `internal` keyword, feel free to close this PR!)

PR #146, opened 2015-12-23
ImageRow with Action Sheet

weyert:
I have changed the implementation of _ImageRow so that it now ask which source type should be used when sourceTypes property has more then one source type defined.

What do you think of it?


danipralea:
can this be checked? It could use it as well :P

PR #159, opened 2016-01-06
Disable autocorrection and autocapitalization in URLRow()

jasonnam:
I think it would be necessary to disable autocorrection and autocapitalisation in URLRow() by default.


mtnbarreto:
Thanks for contributing!

PR #172, opened 2016-01-10
Feature: Typed Row

### LINKED ISSUE #107: "Typed" Row?
How to define a "Typed" row in Eureka? For example if you have a phone input which also have a choosable type like "Private", "iPhone" or "Business"?

    PHONE
    [ Private | 45678 ]
    [ iPhone  |       ]

marbetschar:
Added the basic functionality to support labeled rows. The following works for every existing row by using the `title` as label/type value.

Fixes #107 

The new `displayTitleFor` closure is called whenever the title for the row changes. It allows to display another text for the title as it's value.

Additionally there is a new `titleDidSelect` closure which can be added to a row for custom label tap behaviour - for example to open a viewController for selecting another label.

A full working example:

```
<<< TextRow() {
    $0.title = String(kABWorkLabel) //"_$!<Work>!$_"
    $0.displayTitleFor = {(title: String?) -> String? in
        if title == String(kABWorkLabel) {
            return "Work"
        }
        return nil
    }
}.titleDidSelect{ row in
    UIAlertView(title: "Change Label", message: "A view controller could be used to change the label", delegate: nil, cancelButtonTitle: "Ok").show()
}
```


marbetschar:
@mtnbarreto I'm aware of the "one-commit-per-feature" issue here; but I just wanted to know if there's any chance this pull request (or any other one providing this feature) would make it into the official release any ways?

mtnbarreto:
It seems to me this can be accomplished by setting up a tap gesture recognizer to titleLabel from within `cellSetup` and handling the gesture event accordingly.

PR #175, opened 2016-01-11
TextViewCell Placeholder/TextView Layout

holtat:
Fixed bug where the TextViewCell's placeholderLabel did not line up with the textView's content.

To see an example of the issue, run the Example project and open the Native iOS Event Form. The last row entitled Notes has a placeholder, but when you start typing, the content does not line up with the placeholder's position.

A good writeup of the underlying problem and solution can be found here: http://stackoverflow.com/a/18987810/2712112

Before:
screen shot 2016-01-11 at 9 13 34 am

After:
screen shot 2016-01-11 at 9 13 56 am


mtnbarreto:
:+1:  @holtat Thanks for contributing!

PR #176, opened 2016-01-12
Made arrayData public let

mikaoj:
Changed arrayData property to be a public let.

Backstory: I'm using Eureka in a project where I want to create my own custom selector controller. And when doing that it helps if I can access the arrayData property, as you do with SelectorViewController/MultipleSelectorViewController/SelectorAlertController/etc.


Tramvai:
Awesome. Had the same issue. :+1:

PR #183, opened 2016-01-14
Fix image row source types

ffittschen:
Hi, I really appreciate the refactored `ImageRow` and the use of `ActionSheets`!

Looking at the source, I thought I can create an `ImageRow` like this:
```
<<< ImageRow(){
    $0.title = "ImageRow"
    $0.sourceTypes = [ImageRowSourceTypes.Camera, ImageRowSourceTypes.PhotoLibrary]
}
```
But this doesn't work. Xcode says "Type 'ImageRowSourceTypes' has no member 'Camera'".
I tried to change the options in `ImageRowSourceTypes` to public, but then the app crashes on the simulator with the message "Source type 1 not available".

This error appeared, because if an `OptionSetType` gets initialized with `0` it automatically referes to empty / `.None` (which is the case for `UIImagePickerControllerSourceType.PhotoLibrary`). This produced some weird behaviour, causing the crash described above. 

I changed the code to make the following code possible, fix those issues which caused the crash after making the options public and wrote a few basic tests.

```
<<< ImageRow(){
    $0.title = "ImageRow"
    $0.sourceTypes = .PhotoLibrary
}
<<< ImageRow(){
    $0.title = "ImageRow"
    $0.sourceTypes = [ImageRowSourceTypes.Camera, ImageRowSourceTypes.PhotoLibrary]
}
```

Just let me know if there is any feedback ;)

### DIFF
--- a/Eureka.xcodeproj/project.pbxproj
+++ b/Eureka.xcodeproj/project.pbxproj
@@ -7,6 +7,7 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		048D0DA61C46A7F700399A22 /* ImageRowTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 048D0DA51C46A7F700399A22 /* ImageRowTests.swift */; };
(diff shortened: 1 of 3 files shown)


mtnbarreto:
:+1: @ffittschen Thanks for the fix!

weyert:
Thanks :+1:

PR #184, opened 2016-01-15
Fix a bug that removes existing observers when other type of observer is added

jechol:
### Bug Reproduction
1. Add .Hidden observer
2. Add .Disabled observer
3. Then .Hidden observer is removed.

### Bug Demo
Find out working bug reproduction demo.
https://github.com/jechol/eureka-mixed-hidden-disabled-observer-bug-reproduce


mtnbarreto:
@jechol thank for fixing the issue and for the repository that reproduces the issue. I really appreciated it!

PR #186, opened 2016-01-17
Make ViewController Designated Initializers Public

colinmorelli:
Unless I'm missing something, not declaring this initializers as public makes it extremely painful to subclass these controllers. This PR simply exposes the designated initializer as public, so subclasses can override initializers as necessary.


mtnbarreto:
:+1:

PR #202, opened 2016-01-22
Add teardown() to Cell to clean up delegates and targets.

TeunVR:
The PickerInlineRow always crashes when you close it while it is still spinning.
CountDownInlineRow also, but less often.

Added a teardown() function on the Cell to remove delegates from pickers and cleanup addTarget(self) references.

teardown() is called when the Row is deinitialised, it is the opposite of setup()


mtnbarreto:
@TeunVR Why don't use `deinit`? i would prefer to use `deinit ` instead of teardown.

PR #211, opened 2016-01-26
Fix crash issue:  The section contains InlineRow and FieldRow, expand Inline and select FieldRow's titleLabel will crash.

KevinGong2013:
Fix crash issue: row.baseCell.cellBecomeFirstResponder() may be cause InlineRow collapsed then section count will be changed. Use orignal indexPath will out of  section's bounds.


mtnbarreto:
@KevinGong2013 Thanks for contributing!

KevinGong2013:
^_^

PR #225, opened 2016-02-03
Fix crash during PickerInlineRow collapse

### LINKED ISSUE #208: Crash during collapse of PickerInlineRow
There seems to be an edge case that throws a "nil found exception" when using the PickerInlineRow.  The error can be replicated using the Picker row by swiping the picker dial (so it bounces, beyond the last option), and then immediately closing the inline picker view row.  

My best guess is that the logic for selecting a value in the picker is running AFTER the inline row has been, at least partially, deallocated.  The specific error seems to be that the `pickerRow` property is nil when grabbing the `pickerView:titleForRow:forComponent`.  

Because the bouncing animation is still happening when the cell is collapsed, when the bouncing stops, the cell no longer has the pickerRow available to draw data from.

Let me know if help is needed replicating the issue.

SebastianOsinski:
Fix crash which occurs durring collapse while picker bounces beyond last value. Resolves issue #208

### DIFF
--- a/Source/Cells.swift
+++ b/Source/Cells.swift
@@ -610,7 +610,7 @@ public class PickerCell<T where T: Equatable> : Cell<T>, CellType, UIPickerViewD
         return picker
         }()
     
-    private var pickerRow : _PickerRow<T> { return row as! _PickerRow<T> }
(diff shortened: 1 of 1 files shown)


mtnbarreto:
@SebastianOsinski thanks for contributing!

Issue #72, opened 2015-11-04
ImageRow let access to NSURL for the UIImageView


minuscorp:
It would be great if you add a variable to fill the ```NSURL``` for the ```UIImagePickerDelegate``` info dictionary by the key ```UIImagePickerControllerMediaURL``` or there's actually a way to get that image URL, because is mainly needed to post it to the Internet by a Multipart Form.

Maybe just save that ```UIImage``` into the app contents and upload it?

mtnbarreto:
@minuscorp done!

fornieles:
Hello everyone,

I'm trying to access to imageURL but my code is not working, I should have an image loaded : 
`if let url = NSURL(string: "https://etc..") {`
     ` if let res = NSData(contentsOfURL: url) {`
           `$0.value = UIImage(data: res)`
      `}`
`}`

Do you guys have an idea ? The picture is not loaded.

minuscorp:
I don't get where the problem really is...

Are you trying to get `imageURL` after picking an image from the camera roll or you're expecting `imageURL` to be set after setting by `row.value`?

fornieles:
Oh sorry, I am expecting imageUrl to be set after setting by row.values.

mats-claassen:
That will not happen. You have to wait for the user to pick an image or set it yourself.

bitbibit:
@minuscorp Know how to take the image as a URL value after it's been inputted by a user?
I'm trying to save a user image to firebase 

ERROR is `'InvalidFirebaseData', reason: '(setValue:) Cannot store object of type UIImage at pledgeImg. Can only store objects of type NSNumber, NSString, NSDictionary, and NSArray.'`

Thanks


      
```
// Get a value of the IMAGE
        let specImgRow: ImageRow? = form.rowBy(tag: "specImg")
        let specImg = specImgRow?.value
```


      // Trying to store it into Firebase 
       guard let image = Img
            else { return }

        guard let uploadData = UIImageJPEGRepresentation(image, 0.3)
            else { return }
        

        let fileName = NSUUID().uuidString

        Storage.storage().reference().child("images").child(fileName).putData(uploadData, metadata: nil) { (metadata, err) in
            
            if let err = err {
                print("Failed to upload pledge image",err)
            }
            
            Storage.storage().reference().child("images").child(fileName).downloadURL(completion: { (url, error) in
                if error != nil {
                    print(error as Any)
                } else {
[...]

PR #231, opened 2016-02-04
added imageURL property to ImageRow, it's readOnly and it's set up wh…

mtnbarreto:
…en the user select an image from camera roll.


mtnbarreto:
close #72

PR #235, opened 2016-02-05
fix issue: DecimalRow formatter wrong for too-big number?

su23:
Changing type of DecimalRow from Float to Double


mtnbarreto:
:+1:  @su23 Thanks for contributing!

PR #255, opened 2016-02-11
Adding option to clear ImageRow 

btanner:
if it has a value and a property allowImageClear which is Bool and set to true by default.

This is my second attempt at this pull request. This one should work better than last time.

Wasn't really sure if I had to do my own fork to make this work, but here we are :-)

Learning git and github together.


mtnbarreto:
@btanner great work! Thanks for contributing!

PR #271, opened 2016-02-17
Non developer readme contribution

NathanJPlummer:
Added a couple sections to the README file.  "When to use Eureka!" which describes in simple terms common use case scenarios, and "Non-Technical Introduction" which is meant to explain the purpose of Eureka! to non-developers.


mats-claassen:
Hi @NathanJPlummer , why do you think this is necessary? (as it is not common in other repositories)

We think we could add it but not as the main introduction to Eureka as it is mainly used by developers.

NathanJPlummer:
At the risk of sounding flippant, I'm not completely sure you do.  I'm part of an online programming training course and this was part of an assignment.

That being said, as a new programmer I often check out projects on GitHUb I might find useful for a project, and find the overly technical jargon intimidating to newcomers.  For example, as a new programmer I had no idea what a DSL specification was (now that I know, it's pretty useful).

I actually had a [longer, more detailed version](https://github.com/NathanJPlummer/Eureka/blob/master/README_Non_Technical.MD) you can read at the link.  My project partner felt (and I agreed) it was way too much to add to your README.

If you think you may be interested in a less technical introduction as a seperate file I could do a pull request on that.

mtnbarreto:
@NathanJPlummer I liked your non-technical documentation but i think it's more about DSL than Eureka. 
Eureka documentation is not intended to explain what is a DSL and I would expect that a beginner search into google if it doesn't understand non-eureka related concepts.

PR #283, opened 2016-02-23
Move "How to get the form values" section to "Usage"

scottrhoyt:
Seeing as it is such a common requirement, I though it would be good to move the instructions for retrieving form values to the `Usage` section.


mtnbarreto:
:+1:

PR #286, opened 2016-02-25
Selector row now requires generic cell parameter

mats-claassen:
_Alert and _ActionSheet rows also permit custom cell


weyert:
I noticed a breaking change in 1.4.1 as mentioned in the `CHANGELOG` file only when I updatee the `LocationRow` from the `CustomCells.swift` file it's still not compiling for me.

PR #325, opened 2016-03-09
Allows AlertRow cancel title to be changed. close #323

### LINKED ISSUE #323: How to dynamically translate AlertRow cancel button label ?
Hello,

is there a way to dynamically set another title to the action row cancel option (or any other one) in an AlertRow ?

Regards.

mtnbarreto:


carlosbdw:
It seems like cancel title still can not be changed

PR #333, opened 2016-03-11
Fixed scrolling to first responder with large section footers

dhf:
We had a TextAreaRow in a section with a large footer text. When navigating using the accessory view the tableView scrolled the footer visible instead of just the textview. The changes fix that behavior an the textView is scrolled to the visible area instead. Also transition to invisible new first responders was improved.


mtnbarreto:
@dhf Thanks for contributing! I don't think the actual implementation is wrong. Most common scenario is to have small footers and current implementation works good for this case. 
I've tested your changes and it does not work properly when back accessory navigation view button is tapped.

PR #339, opened 2016-03-16
Store values for each tag in a dictionary and do not calculate it eve…

### LINKED ISSUE #265: evaluateHidden() will be dreadfully slow with large forms
I've run into this before with other platforms, including code I've written myself.

The issue is that to evaluate the visibility of a row, it might need to examine the value of every other row.  For small forms this is negligible, but as the forms get large, it gets disgusting because it grows with the square of the number of rows.

I have almost finished moving my main form to the platform, and creating the form (my code that adds the rows takes about 5 full seconds on an iPad Air 2.  Similarly, setting the values from a dictionary takes about the same time.  So, for me to load up a stored form off disk is a 10 second delay for the user.

Operations within the form (like setting a Yes/No field that shows 20 additional rows) can have a 2-3 second delay as well, which makes it feel very sluggish.

There is a fix!  Simply, don't call form.dictionaryValuesToEvaluatePredicate() ever single time evaluateHidden() is called.

Eureka is a bit too complicated for me to know if generally speaking there is a "starter" time in which we can cache the result of that function and then clear it after the cascade of evaluateHidden is called.  But, in my special case I have written a crappy Proof-Of-Concept hack that seems to work. You will surely be able to point me to a better way.

My hack is this:
1) Add static variables to BaseRow:
- flatDictCache: [String : AnyObject]
- skipHiddenEval: Bool = false
- useDictCache: Bool = false

2) Update evaluateHidden in BaseRow to the following:
```Swift
    public final func evaluateHidden() {
        guard let h = hidden, let form = section?.form else { return }
        switch h {
            case .Function(_ , let callback):
                hiddenCache = callback(form)
            case .Predicate(let predicate):
				if BaseRow.skipHiddenEval{
					hiddenCache=false;
				}else{
					if BaseRow.useCache == false{
						hiddenCache = predicate.evaluateWithObject(self, substitutionVariables: form.dictionaryValuesToEvaluatePredicate())
					}else{
						 hiddenCache = predicate.evaluateWithObject(self, substitutionVariables: BaseRow.flatDictCache)
					}
			}
        }
        if hiddenCache {
            section?.hideRow(self)
        }
        else{
            section?.showRow(self)
        }
    }
```
3) Make dictionaryValuesToEvaluatePredicate() public.

4) In my code, before I create and load my form, I just do:
```Swift
	final func ensureFormFullyLoaded(){
		BaseRow.skipHiddenEval=true;

		if !(self.formFieldsCreated){
			self.makeFormFields();
		}
		
		if !(self.isOnlyAStub){
			self.fullyLoadFile();
		}
		
		BaseRow.skipHiddenEval=false;
		BaseRow.useCache=true;
		BaseRow.flatDictCache =theForm.dictionaryValuesToEvaluatePredicate();
		for thisRow:BaseRow in theForm.allRows{
			thisRow.evaluateHidden();
		}
		BaseRow.useCache=false;
}
```

This is totally a gross hack. One way to get most of the value of this into Eureka naturally might be to:
1) Store flatDictCache in the Form class as a private member.
2) Set value of flatDictCache on the every call to evaluateHidden if flatDictCache is nil.
3) Set flatDictCache to nil whenever the form changes (value changed or row/section added/removed)

This will speed up the form building stage. Every time you add a new row you'll only calculate flatDictCache's content once, not once per existing row in the table.

It will also speed up the runtime case where value is set that requires evaluateHidden to be triggered, again b/c it'll be calculated once instead of once per affected row.

Is there any easy way to track down "every time the form changes" to clear that cache? Maybe I could work on a patch if I knew that.

In the meantime, my quick hack has taken my form loading time from 10000 ms to 100 ms.

mats-claassen:
…ry time we evaluateHidden

Fixes #265

### DIFF
--- a/Eureka.xcodeproj/project.pbxproj
+++ b/Eureka.xcodeproj/project.pbxproj
@@ -181,24 +181,24 @@
 		2859CD901C7CF1FD0002982F /* Core */ = {
 			isa = PBXGroup;
 			children = (
-				28EEBFF71C7E241200300699 /* SelectableRowType.swift */,
+				2859CDE81C7DD78E0002982F /* BaseRow.swift */,
+				28EEBFFC1C7E25F100300699 /* Cell.swift */,
+				28EEBFE41C7E1B3B00300699 /* CellType.swift */,
+				2859CD921C7CF1FD0002982F /* Core.swift */,
+				2859CDBD1C7D137D0002982F /* Form.swift */,
+				2859CDC31C7D19C50002982F /* HeaderFooterView.swift */,
+				2859CD931C7CF1FD0002982F /* Helpers.swift */,
 				28EEBFF51C7E240B00300699 /* InlineRowType.swift */,
+				2859CDC51C7D1A2E0002982F /* NavigationAccessoryView.swift */,
+				28EEBFFA1C7E25C500300699 /* Operators.swift */,
+				28EEBFEB1C7E224200300699 /* PresenterRowType.swift */,
+				28EEBFEF1C7E23F400300699 /* Row.swift */,
+				28EEBFED1C7E231C00300699 /* RowControllerType.swift */,
 				28EEBFF31C7E240000300699 /* RowProtocols.swift */,
 				28EEBFF11C7E23FA00300699 /* RowType.swift */,
-				28EEBFEF1C7E23F400300699 /* Row.swift */,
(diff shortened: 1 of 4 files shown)


btanner:
Thanks!  I updated my fork from upstream to get the XCode 7.3 updates and your fix here resolved my slowness.

PR #370, opened 2016-04-02
Update rows statuses of "hidden" and "disabled" after adding all rows

su23:
If you have some complex dependency between rows (for example in row A visibility is depended on the value of row B which is located below row A) then you need to have an access to this row (which may not exists in the UI at that moment). Thus, the "visibility" or "disable" status of row A is not right.

In my change I split the logic of row adding and updating phases. Thus, in the first step we add all rows to the section and only after that evaluate "hidden" and "disable" rules for each added row.


mats-claassen:
Hi @su23 I think this does not fix the real problem. Suppose the following case:
You add two rows that depend on a third. And after that you add the third in a new call.
In this case you would not reevaluate the previous rows. We have to find a better solution.

PR #401, opened 2016-04-29
Added onCellSelection example

rcrsv2n:
See issue #3. There are only examples of using buttons with segues and it is not immediately clear to use the onCellSelection function.

### DIFF
--- a/Example/Example/ViewController.swift
+++ b/Example/Example/ViewController.swift
@@ -40,7 +40,7 @@ class HomeViewController : FormViewController {
         
         form  +++=
             
-            Section(footer: "These are 10 ButtonRow rows") {
+            Section() {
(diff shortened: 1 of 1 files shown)


Example/Example/ViewController.swift
```diff
@@ -92,9 +92,24 @@ class HomeViewController : FormViewController {
                     row.title = row.tag
                     row.presentationMode = .SegueName(segueName: "ListSectionsControllerSegue", completionCallback: nil)
                 }
+        +++ Section()
+                <<< ButtonRow() { (row: ButtonRow) -> Void in
+                   row.title = "About"
+                }  .onCellSelection({ (cell, row) in
```
mtnbarreto: @rcrsv2n i think you missed a [weak self] here

Issue #405, opened 2016-05-05
Smarter table view row deselection


ZevEisenberg:
Before submitting issues ...

The code in question is here, in `FormViewController` from `Core.swift`. I'm looking at the version on `master` at the time of this writing, which is from commit https://github.com/xmartlabs/Eureka/commit/826d8d9b75b0f3d8890c839360c4f93d7dbb92e8

```swift
public override func viewWillAppear(animated: Bool) {
    super.viewWillAppear(animated)
    if let selectedIndexPath = tableView?.indexPathForSelectedRow {
        tableView?.reloadRowsAtIndexPaths([selectedIndexPath], withRowAnimation: .None)
        tableView?.selectRowAtIndexPath(selectedIndexPath, animated: false, scrollPosition: .None)
        tableView?.deselectRowAtIndexPath(selectedIndexPath, animated: true)
    }
    NSNotificationCenter.defaultCenter().addObserver(self, selector: #selector(FormViewController.keyboardWillShow(_:)), name: UIKeyboardWillShowNotification, object: nil)
    NSNotificationCenter.defaultCenter().addObserver(self, selector: #selector(FormViewController.keyboardWillHide(_:)), name: UIKeyboardWillHideNotification, object: nil)
}
```

The problem with this is that if you use an interactive transition to start dismissing the pushed view (so that `self` is reappearing), and then change your mind and cancel the dismissal, the row is still deselected, a subsequent "Back" navigation doesn't show the row deselecting again.

The solution is to use the trick that I outline here: https://gist.github.com/ZevEisenberg/8951478e8c6907bd5e5a

The idea is that you use the transition coordinator to find out if the animation was canceled, and if it was, it reselects the desired rows so that they will still be selected on subsequent dismissal attempts.

ZevEisenberg:
My solution also assumes that you're being dismissed in a navigation controller, which may not be the case always.

mtnbarreto:
@ZevEisenberg Thanks for reporting the issue and provide info to solve it. It seems related with #402. I will look into it and get back to you shortly.

mtnbarreto:
I have not tested in the simulator neither a real device but it seems to me it should work for any transition even custom transitions. Why did you say "My gist also assumes that you're being dismissed in a navigation controller" ?

ZevEisenberg:
[...]

PR #406, opened 2016-05-05
Better row deselection behavior during interactive transitions

ZevEisenberg:
Closes https://github.com/xmartlabs/Eureka/issues/405

### DIFF
--- a/Source/Core/Core.swift
+++ b/Source/Core/Core.swift
@@ -481,11 +481,36 @@ public class FormViewController : UIViewController, FormViewControllerProtocol {
     
     public override func viewWillAppear(animated: Bool) {
         super.viewWillAppear(animated)
-        if let selectedIndexPath = tableView?.indexPathForSelectedRow {
-            tableView?.reloadRowsAtIndexPaths([selectedIndexPath], withRowAnimation: .None)
-            tableView?.selectRowAtIndexPath(selectedIndexPath, animated: false, scrollPosition: .None)
-            tableView?.deselectRowAtIndexPath(selectedIndexPath, animated: true)
+
+        let selectedIndexPaths = tableView?.indexPathsForSelectedRows ?? []
+        tableView?.reloadRowsAtIndexPaths(selectedIndexPaths, withRowAnimation: .None)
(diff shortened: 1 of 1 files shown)


Source/Core/Core.swift
```diff
@@ -481,11 +481,36 @@ public class FormViewController : UIViewController, FormViewControllerProtocol {
...
-            tableView?.selectRowAtIndexPath(selectedIndexPath, animated: false, scrollPosition: .None)
-            tableView?.deselectRowAtIndexPath(selectedIndexPath, animated: true)
+
+        let selectedIndexPaths = tableView?.indexPathsForSelectedRows ?? []
+        tableView?.reloadRowsAtIndexPaths(selectedIndexPaths, withRowAnimation: .None)
+        selectedIndexPaths.forEach {
+            tableView?.selectRowAtIndexPath($0, animated: false, scrollPosition: .None)
+        }
+
+        let deselectionAnimation = { (context: UIViewControllerTransitionCoordinatorContext) in
```
mtnbarreto: @ZevEisenberg `[weak self]` should be added in the closure, right? same applies to `reselection ` closure. I can make these changes if you agree on them.

ZevEisenberg: @mtnbarreto no need to `[weak self]`, since you're in an animation block that `self` is not retaining. Only need `[weak self]` when the thing that the block is capturing (and therefore retaining) is also retaining the block. Here, the animation system is holding the block, and will eventually release it, which will then release the captured `self`.

mtnbarreto: Got it. I'm not completely sure how swift manage these references and which one is released first. I've already added the `[weak self]`, the result should be the same. Note that i didn't add it to foreach since it explicitly declare `@noescape`.

ZevEisenberg: Fair enough. A good rule of thumb is that if it's an animation block, and it's going to run once, then it's fine to not use `[weak self]`. If it's a repeating animation, more care must be taken, because the animation system will retain the block indefinitely, so you can't rely on canceling the animation in `deinit`.

PR #412, opened 2016-05-07
DecimalRows uses DecimalFormatter(). No need to initialize with NSNumberFormatter().

p-sun:
All DecimalRows in the Example project use DecimalFormatter(), so initializing with NSNumberFormatter() here is redundant.


mats-claassen:
Thanks for contributing but this should still be the default configuration. The example is just showing how we can change this

PR #413, opened 2016-05-08
Extended the functionality of PhoneRow to format phone numbers.

p-sun:
Added a PhoneFormatter class that formats phone numbers in PhoneRows, a feature that was previously unimplemented. A sample row was added to the Example project.

Phone numbers in PhoneRow are formatted as follows:
1 (514) 123-5234
(514) 123-5234
(514) 123-4


SASDOE:
@p-sun How does this handle numbers that aren't american? Is there anyway you could use this https://github.com/marmelroy/PhoneNumberKit?

p-sun:
@SASDOE That's a fantastic idea! I attempted to use the PhoneNumberKit that you suggested to re-implement PhoneRow in #436. It seems to work when I run it on my device, but I'm a bit lost on why it's failing the check.

 Yes, #413 is a lightweight solution that only formats in the American format.

SASDOE:
Anyone want to integrate this?

mats-claassen:
First of all it has conflicts with master branch.
Second, we will soon move to Swift 3 so this should be implemented in Swift 3 and merged onto Swift3 branch.
And I think it would be good to have some basic test for this formatter.

mtnbarreto:
Let's implement a phone input with formatters using https://github.com/marmelroy/PhoneNumberKit but i would like to provide it as custom row extension. 
Anyone wants to implement and add it to https://github.com/eurekacommunity ??

PR #418, opened 2016-05-10
Added a row that takes a String link, and opens the link in Safari. 

p-sun:
Clicking on this custom row opens the link in Safari. I've also included a CustomWebViewController to include the choice of using an in-app web view instead of Safari.


mats-claassen:
Hi @p-sun we think this is a very custom implementation and there might be other implementations. We also do not want to mix too many particular features into the examples as they should show more about Eureka features

PR #455, opened 2016-05-25
Ability to find the current FormViewController as a childViewControll…

Zigzag968:
Ability to find the current FormViewController as a childViewController of a ViewController in formViewController() method (Cell.swift)

I had to do it because of the structure of my view controllers that requires FormViewController to be a child view controller. Indeed, the RowNavigationOptions were missing, and the cellHighlight/Unhighlight were never called, as the formViewController method can't find the FormViewController as a responder in the chain


mtnbarreto:
i'm gonna fix it.

Issue #495, opened 2016-06-13
MultipleSelectorRow with key value list how to show values on presented viewcontroller


avalla:
related issue #327 

I cannot understand how to use a key value dropdown list with MultipleSelectorRow, so I'm opening an issue because I don't know where to ask.

The presented view controller shows the keys, but I want to show the values

here is my code:

```swift
let countries = [ "uk" : "United Kingdom", "fr" : "France", "us" : "United States"]
form +++ Section("Required fields".uppercaseString)
	<<< MultipleSelectorRow<String>("countries") { row in
	row.title = "Countries"
	row.options = countries.map { $0.0 }
	// Show the labels of selected values
	row.displayValueFor  = {
		if let t = $0 {
			return t.map({ countries[$0]! }).joinWithSeparator(", ")
		}
		return nil
	}
	row.onPresent({ from, to in
		// Add the label of the value on the right side of the cell after selection
		to.selectableRowCellUpdate = { cell, row in
			row.displayValueFor = { val in
				if let value = val {
					return countries[value]
				}
				return nil
			}
		}
	})
}
```
- Environment: Eureka 1.6.0, Xcode 7.3.1, iOS 9.3

Thanks!!

![screenshot 2016-06-07 12 45 13](https://cloud.githubusercontent.com/assets/986614/15855102/f3cf4080-2cad-11e6-8329-9e8bb6342c96.png)
![screenshot 2016-06-07 12 45 19](https://cloud.githubusercontent.com/assets/986614/15855108/f7dbece6-2cad-11e6-920b-f8214690d0b6.png)
![screenshot 2016-06-07 12 45 25](https://cloud.githubusercontent.com/assets/986614/15855112/fcd891ea-2cad-11e6-829c-496d9cab9236.png)

avalla:
Fixed with pull request #511 

```swift
      let countries = [ "uk" : "United Kingdom", "fr" : "France", "us" : "United States"]
      form +++ Section("Required fields")
        <<< MultipleSelectorRow<String>("countries") { row in
          row.title = "Countries"
          row.options = countries.map { $0.0 }
          row.onPresent({ from, to in
            // Decode the value in row title
            to.selectableRowCellSetup = { cell, row in
              if let value = row.selectableValue {
                row.title = countries[value]
              }
            }
          })
      }
```

PR #511, opened 2016-06-22
Added cell setup callback in MultipleSelectorViewController

avalla:
I needed to change the value of title of the row in a MultipleSelectorViewController, but there was no way to do it.

So I addeed a new callback to do this.

Related issue #495

### DIFF
--- a/Source/Rows/Controllers/MultipleSelectorViewController.swift
+++ b/Source/Rows/Controllers/MultipleSelectorViewController.swift
@@ -15,6 +15,7 @@ public class _MultipleSelectorViewController<T:Hashable, Row: SelectableRowType 
     /// The row that pushed or presented this controller
     public var row: RowOf<Set<T>>!
     
+    public var selectableRowCellSetup: ((cell: Row.Cell, row: Row) -> ())?
     public var selectableRowCellUpdate: ((cell: Row.Cell, row: Row) -> ())?
 
     /// A closure to be called when the controller disappears.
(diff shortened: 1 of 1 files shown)


mtnbarreto:
Hi @avalla  Why don't you use `selectableRowCellUpdate `?

avalla:
I tried to use the update callback, but it is fired just after scrolling, the first time that the cell is displayed it wasn't fired.

it is not better to manipulate the title of the cell when during setup?

maybe i was using selectableRowCellUpdate it in the wrong way, this was my test:

```swift
let countries = [ "uk" : "United Kingdom", "fr" : "France", "us" : "United States"]
form +++ Section("Required fields".uppercaseString)
	<<< MultipleSelectorRow<String>("countries") { row in
	row.title = "Countries"
	row.options = countries.map { $0.0 }
	// Show the labels of selected values
	row.displayValueFor  = {
		if let t = $0 {
			return t.map({ countries[$0]! }).joinWithSeparator(", ")
		}
		return nil
	}
	row.onPresent({ from, to in
		// Add the label of the value on the right side of the cell after selection
		to.selectableRowCellUpdate = { cell, row in
			row.displayValueFor = { val in
				if let value = val {
					return countries[value]
				}
				return nil
			}
		}
	})
}
```

mtnbarreto:
Yeah, the `displayValueFor ` should be changed before cell is displayed in the table view since this callback is used to update the cell text. 

As a workaround you can set up the textLabel.text from within the `selectableRowCellUpdate` as code below shows...

```swift
to.selectableRowCellUpdate = { cell, row in
            row.displayValueFor = { val in
                if let value = val {
                    return countries[value]
                }
                return nil
            }
            // don't remember exactly if we have to set up textLabel or detailTextLabel. 
            cell.textLabel.text = row.displayValueFor(row.value) // < ------------------------
        }
```

Anyway, the callback you added makes sense.
Thanks for contributing!

avalla:
thanks to your for this useful library!! 👍

PR #548, opened 2016-07-09
Converts to Swift 3 Syntax (2 Test Failures)

slammer8:
## What it does

- Fully converts to Swift 3 syntax for framework, project, and tests.

## What it doesn't do

Pass all the tests. There are two test failures:
1. CollectionTests - `testDelegateFunctions()`
2. HiddenRowTests - `testHiddenSections()`

## How to test

Run the project, run the tests.

## Notes

Don't understand the Eureka codebase enough, so I didn't want to try to fix the tests.

### DIFF
--- a/Example/Example/ViewController.swift
+++ b/Example/Example/ViewController.swift
@@ -585,16 +585,16 @@ class NativeEventFormViewController : FormViewController {
                     let endDate: DateTimeInlineRow! = self?.form.rowByTag("Ends")
                     
                     if row.value ?? false {
-                        startDate.dateFormatter?.dateStyle = .medium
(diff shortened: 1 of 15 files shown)


charlymr:
Same here. Tried and debug but not sure why those 2 particulars test are failing.
Reverting to the Author @mats-claassen  for that.

mats-claassen:
@slammer8 Which Xcode beta version are you using?
I noticed you changed some enums (like from `none` to `noStyle`, for example). But in fact it seems that the old ones are the correct ones as to Xcode beta 2 which is the newest since July 5.

slammer8:
@mats-claassen, you're correct- updating to beta 2 now. Thanks for pointing that out!

PR #551, opened 2016-07-11
Deselection changes

mats-claassen:
onSelectSelectableRow will not be called when nothing changes
 (when `enableDeselection = false` and user deselects a row)

Added enableDeselection customisation parameter to SelectorViewController

### DIFF
--- a/Example/Example/ViewController.swift
+++ b/Example/Example/ViewController.swift
@@ -668,7 +668,9 @@ class NativeEventFormViewController : FormViewController {
                     $0.title = $0.tag
                     $0.options = RepeatInterval.allValues
                     $0.value = .Never
-                }
+                }.onPresent({ (_, vc) in
+                    vc.enableDeselection = false
+                })
         
         form +++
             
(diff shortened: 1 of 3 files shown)


AleksandarPetrov:
Consider the following scenario. I have a form with one SelectableSection. I don't want to enable deselection. Then I'd like to navigate to the next screen if option is selected. However when I get back the option is already selected and I'm not able to select it second time. Neither `onSelectSelectableRow` nor `valueHasBeenChanged(for row: BaseRow, oldValue: Any?, newValue: Any?)` are called. I think it's ok to not call `valueHasBeenChanged`, because in fact nothing is changed, however I do expect `onSelectSelectableRow` to be called because row is actually selected. because a check in your `prepare(selectableRows rows: [BaseRow])` method

```
// Check if row is not already selected
if enableDeselection || row.value == nil {
    row.value = row.value == nil ? row.selectableValue : nil
    row.updateCell()
    s.onSelectSelectableRow?(cell, row)
}
```

I found a workaround:

PR #585, opened 2016-07-25
Fixed release configuration

mats-claassen:
Also adapted to new Xcode beta 3.
Still several warnings to be addressed in future.
Not sure why `subPredicates` in `Predicate` are AnyObject and not Predicate's...


alpennec:
Will you merge to Swift 2.3 branch also?

PR #592, opened 2016-07-28
fix #566

### LINKED ISSUE #566: When I insert a new section ,the section is strange
When I insert a new section,I found two strange thing.

*  The section header title will be upper,I set it as "New Section Optional(1))",but the section header view is "NEW SECTION OPTIONAL(1)".
*  When i add new section one by one,the first section will be slide in from the left one by one.

Xcode 7.3.1  iPhone 6 plus 9.3.2

			.onCellSelection({ (cell, row) in
				let sectionTitle = "New Section ".stringByAppendingString("\(row.indexPath()?.section))")
				let section = Section(sectionTitle)
				print(sectionTitle)
				self.form.insert(section, atIndex: (row.indexPath()?.section)!)
		})

You can download a demo from https://github.com/sunbohong/Eureka/tree/demo

sunbohong:


mtnbarreto:
thanks for pointing out the issue. Now it's already fixed in master branch.

PR #599, opened 2016-08-02
Fix: displaying new value of slider with displayValueFor

svedm:
displayValueFor not worked for updates of sliderRow


mtnbarreto:
👍

PR #604, opened 2016-08-10
Update for Xcode 8 beta 5

jyunderwood:
This PR gets the swift3 branch running with tests passing under Xcode 8 beta 5 but seem to time out on travis.

One issue I ran in to is `canBecomeFirstResponder` is now a [read-only property](https://developer.apple.com/reference/uikit/uiresponder/1621130-canbecomefirstresponder
) instead of a method in the latest release. I have commented out the two overrides in the project with a TODO comment.

### DIFF
--- a/Eureka.playground/Contents.swift
+++ b/Eureka.playground/Contents.swift
@@ -55,10 +55,10 @@ formController.view
 formController.form +++ Section("Callbacks") <<< SwitchRow("scr1"){ $0.title = "Switch to turn red"; $0.value = false }
     .onChange({ row in
         if row.value == true {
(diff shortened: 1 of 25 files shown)


mats-claassen:
Travis integration will probably always fail as Travis is currently using Xcode beta 4. 
Can we override `canBecomeFirstResponder` with a custom getter?

jyunderwood:
@mats-claassen Yes, they they are now replaced with computed getter overrides.

mtnbarreto:
👍

mtnbarreto:
thanks for the great work @jyunderwood !

PR #615, opened 2016-08-18
Swift3

Legoless:
Updates for Beta 6 support. Currently it compiles on Xcode8-Beta6, but some crashes still need to be fixed with number and currency converters.


Legoless:
Yeah, I'm going through a couple of issues now, apparently Fix-it is not always that great.

mtnbarreto:
@Legoless Great work!!, could you please squash these commits into only one commit? so we keep the commit history cleaner.

Legoless:
Sure, I'll open another PR.

PR #654, opened 2016-09-13
Remove PostalAddressRow from project

mats-claassen:
Also moved `navigationAccessoryView` instantiation to `viewWillAppear` because using Eureka from Carthage broke with a lazy instantiation.

PostalAddressRow should soon be found on [EurekaCommunity](https://github.com/EurekaCommunity)


mtnbarreto:
Great work! I would have preferred 2 different pull requests, it's easier to keep track of the changes and accept part of the changes.

PR #658, opened 2016-09-14
Eureka build-in validation feature

mtnbarreto:
@xmartlabs/ios someone can review this? tests will be added shortly.


Source/Core/Row.swift
```diff
@@ -50,45 +49,90 @@ open class RowOf<T: Equatable>: BaseRow {
             }
         }
     }
-    
+
+    /// The typed value of this row.
+    public var value : T?{
+        set (newValue){
+            _value = newValue
+            if validationOptions.contains(.validatesOnChange) || (blurred && validationOptions.contains(.validatesOnChangeAfterBlurred)) || !isValid   {
```
mats-claassen: Why the `|| !isValid`

Source/Core/Row.swift
```diff
@@ -50,45 +49,90 @@ open class RowOf<T: Equatable>: BaseRow {
...
+
+    @discardableResult
+    public override func validate() -> [ValidationError] {
+        validationErrors = rules.flatMap { $0.validateFn(value) }
+        return validationErrors
+    }
+
+    public func addRule<Rule: RuleType>(rule: Rule) where T == Rule.RowValueType{
+        let validFn: ((T?) -> ValidationError?) = { [weak self] (val: T?) in
+            guard let _ = self else { return nil }
```
mats-claassen: Why retaining reference to self here?

Source/Validations/RuleURL.swift
```diff
@@ -0,0 +1,44 @@
...
+public struct RuleURL: RuleType {
+    
+    public init() {}
+    
+    public var id: String?
+    public var validationError = ValidationError(msg: "Field value must be an URL!")
+    
+    public func isValid(value: URL?) -> ValidationError? {
+        guard let value = value else  { return validationError }
+        let regEx = "((https|http)://)((\\w|-)+)(([.]|[/])((\\w|-)+))+"
```
mats-claassen: Why do you not use `RegExpPattern.URL`

Source/Core/Form.swift
```diff
@@ -346,3 +346,17 @@ extension Form {
         kvoWrapper.sections.insert(section, at: formIndex == NSNotFound ? 0 : formIndex + 1 )
     }
 }
+
+extension Form {
+    
+    @discardableResult
+    public func validates() -> [ValidationError] {
```
mtnbarreto: i added the annotation because we may want to run validations and discard the errors since `row.validationErrors` holds it. 

```
let errors: [ValidationError] = row.validationErrors
````

Source/Core/Row.swift
mtnbarreto: if the row is invalid we want to run validations on each value change to show the validation passing as soon as possible to the user. 
We've talked about it before.

PR #666, opened 2016-09-16
Declare some more classes, functions and variables as open

### LINKED ISSUE #643: Permissions Issues for Overriding and Inherit
I am customizing the `ButtonRow`, and when I tried to override the method `prepareForSegue` from the class `BaseRow` and the methods ` customDidSelect` and `customUpdateCell` from the class `Row` I faced the following error _"Overriding non-open instance method outside of its defining module"_. My way around was replacing the access control of these methods from `public` to `open`, which is [now](https://swifting.io/blog/2016/08/17/22-swift-3-access-control-beta-6/) the access level that allows classes to be subclassed and methods overridden.

I'd like to avoid modifying such core classes. Could you guys please change these methods to open, or maybe did you intend these to be public? 

Thanks in advance!

### LINKED ISSUE #646: Can we `open` some class&funcation to make it possible for override and Inherit ?
Thanks for your Eureka!

When I convert my code to Swift 3 version,I found that many class was `public` but not `open`.

I need a cell to show a big image.

If I create it from zero to one,it will cost me a lot of time.

So,I override and Inherit some class to complete it.

But,this code need `open` some class&funcation.

So,can I make a PR for this purpose?

```
final public class BigPushSelectorCell: PushSelectorCell<UIImage>  {

    lazy public var avatarImageView: UIImageView = {
        let avatarImageView = UIImageView()
        avatarImageView.clipsToBounds = true
        return avatarImageView
    }()

    public lazy var placeholderImage: UIImage = {
        let v = UILabel()
        v.backgroundColor = UIColor.white
        v.textColor = UIColor(white: 0.6, alpha: 1.0)
        v.text = self.row.title
        v.sizeToFit()
        return v.snapshotImage() ?? UIImage()
    }()

    required public init(style: UITableViewCellStyle, reuseIdentifier: String?) {
        self.textLabel?.isHidden = true
    }

    required public init?(coder aDecoder: NSCoder) {
        fatalError("init(coder:) has not been implemented")
    }

    public override func setup() {
        height = { self.contentView.bounds.width }
        row.title = nil
        super.setup()
        contentView.addSubview(avatarImageView)
    }

    public override func update() {
        super.update()
        accessoryType = .none
        selectionStyle = .none
        accessoryType = .none

        if let image = row.value {
            avatarImageView.image = image
            avatarImageView.contentMode = .scaleAspectFit
        }
        else {
            avatarImageView.image = placeholderImage
            avatarImageView.contentMode = .center

        }
    }

    override public func layoutSubviews() {
        super.layoutSubviews()
        avatarImageView.frame = contentView.bounds.insetBy(dx: 20, dy: 20)
    }

}

final public class BigImageRow:  _ImageRow<BigPushSelectorCell>, RowType {
    public required init(tag: String?) {
        super.init(tag: tag)
    }

    open override func customUpdateCell() {
        super.customUpdateCell()
        cell.accessoryView = nil
    }
}
```

mats-claassen:
fixes #646 
fixes #643 

@xmartlabs/eureka

### DIFF
--- a/Source/Core/BaseRow.swift
+++ b/Source/Core/BaseRow.swift
@@ -105,12 +105,12 @@ open class BaseRow : BaseRowType {
     /**
      Method that reloads the cell
      */
-    public func updateCell() {}
+    open func updateCell() {}
     
     /**
      Method called when the cell belonging to this row was selected. Must call the corresponding method in its cell.
      */
-    public func didSelect() {}
+    open func didSelect() {}
     
     open func prepareForSegue(_ segue: UIStoryboardSegue) {}
     
@@ -137,7 +137,7 @@ extension BaseRow {
     /**
      Evaluates if the row should be hidden or not and updates the form accordingly
      */
-    public final func evaluateHidden() {
+    public func evaluateHidden() {
         guard let h = hidden, let form = section?.form else { return }
         switch h {
         case .function(_ , let callback):
@@ -156,7 +156,7 @@ extension BaseRow {
     /**
      Evaluates if the row should be disabled or not and updates it accordingly
      */
-    public final func evaluateDisabled() {
+    public func evaluateDisabled() {
         guard let d = disabled, let form = section?.form else { return }
         switch d {
         case .function(_ , let callback):
--- a/Source/Core/Core.swift
+++ b/Source/Core/Core.swift
@@ -549,7 +549,7 @@ open class FormViewController : UIViewController, FormViewControllerProtocol {
     /**
     Called when a cell becomes first responder
     */
-    public final func beginEditing<T:Equatable>(_ cell: Cell<T>) {
+    public func beginEditing<T:Equatable>(_ cell: Cell<T>) {
         cell.row.isHighlighted = true
         cell.row.updateCell()
         RowDefaults.onCellHighlightChanged["\(type(of: self))"]?(cell, cell.row)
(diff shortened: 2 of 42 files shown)


Source/Core/BaseRow.swift
```diff
@@ -137,7 +137,7 @@ extension BaseRow {
     /**
      Evaluates if the row should be hidden or not and updates the form accordingly
      */
-    public final func evaluateHidden() {
+    public func evaluateHidden() {
```
mtnbarreto: keep final

Source/Core/BaseRow.swift
```diff
@@ -156,7 +156,7 @@ extension BaseRow {
     /**
      Evaluates if the row should be disabled or not and updates it accordingly
      */
-    public final func evaluateDisabled() {
+    public func evaluateDisabled() {
```
mtnbarreto: keep final please

Source/Core/Core.swift
```diff
@@ -549,7 +549,7 @@ open class FormViewController : UIViewController, FormViewControllerProtocol {
     /**
     Called when a cell becomes first responder
     */
-    public final func beginEditing<T:Equatable>(_ cell: Cell<T>) {
+    public func beginEditing<T:Equatable>(_ cell: Cell<T>) {
```
mtnbarreto: here too

Source/Core/Core.swift
```diff
@@ -567,7 +567,7 @@ open class FormViewController : UIViewController, FormViewControllerProtocol {
     /**
      Called when a cell resigns first responder
      */
-    public final func endEditing<T:Equatable>(_ cell: Cell<T>) {
+    public func endEditing<T:Equatable>(_ cell: Cell<T>) {
```
mtnbarreto: too

rigobertomolina:
This is an evil pull request lmao

mtnbarreto:
hahaha. it was merged by me 😱

PR #676, opened 2016-09-22
Validation Initializer

kennedykok:
As with the default initializer above, if you want a public structure
type to be initializable with a memberwise initializer when used in
another module, you must provide a public memberwise initializer
yourself as part of the type’s definition.


mtnbarreto:
I have already fixed it!  thanks!

PR #681, opened 2016-09-26
Changing access level of keyboardWillShow and keyboardWillHide methods

svedm:
In swift2 version of eureka we can override this methods for custom logic. This pull request returns those opportunities.


mtnbarreto:
👍

PR #694, opened 2016-10-01
Add unserializing and setting up rows from a dictionary.

amosavian:
This pull is partial, only implements unserializing initializers for Row types from Dictionary. I have to add similar ability to Section and FormViewController later to finally able to setup a form from json/xml.
Also some methods should be added to Button and PushRow to allow interacting with web servers.
Our final goal here is to be able to use this library to create a form which is can be initialized from server and send back results to server.

### DIFF
--- a/Source/Core/Helpers.swift
+++ b/Source/Core/Helpers.swift
@@ -79,3 +79,26 @@ extension NSExpression {
         }
     }
 }
+
+extension UIColor {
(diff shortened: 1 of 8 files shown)


amosavian:
@jessedc any idea?

mstralka:
This could be very useful. Are you still working on it?

amosavian:
@mstralka I can if it's applicable in this repo.

PR #696, opened 2016-10-03
Convert ProvisioningStyle to Automatic

raven:
Provisioning style 'Automatic' is the most consumer friendly provisioning style for frameworks. It permits the host project to dictate the provisioning style. 
Otherwise if a host project specifies Manual provisioning style, xcodebuild will fail when attempting to compile the subproject with: 
> Eureka does not support provisioning profiles. Eureka does not support provisioning profiles, but provisioning profile <defined provisioning profile> has been manually specified. Set the provisioning profile value to "Automatic" in the build settings editor.


mtnbarreto:
👍

PR #719, opened 2016-10-12
Update SelectionType usages in README to Swift 3

mokagio:
Love the project. Glad I could find a way to help, even if so tiny. 💖


mtnbarreto:
@mokagio  Thanks for contributing!

PR #733, opened 2016-10-16
Customizable validation message

ogidow:
This PR becomes validation message is customizable

example
```swift
row.add(rule: RuleRequired("customize message"))
```

### DIFF
--- a/Source/Validations/RuleEmail.swift
+++ b/Source/Validations/RuleEmail.swift
@@ -26,9 +26,9 @@ import Foundation
 
 public class RuleEmail: RuleRegExp {
     
-    public init() {
+    public init(msg: String = "Field value should be a valid email!") {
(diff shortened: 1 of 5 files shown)


mtnbarreto:
code looks good! Thanks for contributing!

Tibb:
Would love to use like this. Unfortunately, it doesn't seemed to work on my code:
capture d ecran 2017-02-09 a 10 24 30

As long as I try to put something else than the basic constructor `RuleRequired()`, I get this _Ruleset_ error.
What am I doing wrong?

allenhuang:
@Tibb 
You need to code like this

`$0.add(rule: RuleRequired(msg:"My Error Message"))`

Tibb:
Thanks @allenhuang , but this issue was actually related to missing imports in a previous version of the library. This is now fixed ;)

allenhuang:
@Tibb 
hmmm, strange. I use the latest version (3.0.0) and still encounter the error you posted.

PR #735, opened 2016-10-17
Explain subclassing of cells with nib better

### LINKED ISSUE #734: "Subclassing cells using the same row" documentation is vague
It seems likely that a common scenario would be to customise the look of a row with a nib file, but retain all of the built-in functionality - this is called out in an explicit section [Subclassing cells using the same row](https://github.com/xmartlabs/Eureka#subclassing-cells-using-the-same-row). 

However, I’ve had no luck following the instructions here. Some issues I’ve managed to resolve after significant investigation, some I haven’t, and I’ve ended up having to just do a full custom row. An example would be super-helpful.

Some of the issues I found were:

* Not really clear what gets subclassed and what doesn’t. I assume from the title of the section, you subclass `TextCell` but not `TextRow` (as it’s final) or `_TextRow`.
* Once you’ve done that, there’s no instructions for how to configure which nib to use. After a fair bit of digging, I found `CellProvider` which I’m setting up in each row config in the form definition - I couldn’t find a sensible way to configure this globally.
* `CellProvider` will throw exceptions every time it’s used until you look at the source and work out it’s trying to load the nib file from the bundle containing the Cell superclass (i.e. Eureka.framework). You’ll always need to explicitly pass a bundle to `CellProvider` for this to work, which is inconsistent with how UIKit nil bundle parameters normally work.
* Once you’ve subclassed the cell and defined a nib, you probably want to wire up some outlets, e.g. to `titleLabel` & `textField`. You can’t. These aren’t defined as `@IBOutlet` in the superclass, some are non-optional, some are get-only; it’s very difficult to get everything properly linked. Best option is to use temporary outlets and shuffle things around in `awakeFromNib`
* Then I got the painful `unknown class in Interface Builder` error, and had to fiddle around with initialisers to get everything loading properly.
* Then once it was loaded, not only was the row height incorrect (breaking the layout), but the superclass busily overwrote a bunch of views and I ended up with an unusable cell.
* I attempted to override the `setup` and `update` methods to prevent the superclass doing its own drawing, but all that gained me was runtime constraint format exceptions.

At this point I gave up. I’m sure there’s some way to get this working, but I couldn’t find it. I’d probably suggest this section is removed from the readme as it doesn’t seem to be a particularly usable extensibility point.

mats-claassen:
closes #734

### DIFF
--- a/README.md
+++ b/README.md
@@ -570,12 +570,24 @@ You can place your own UIViewController instead of `SelectorViewController<T>` a
 ### Subclassing cells using the same row
 
 Sometimes we want to change the UI look of one of our rows but without changing the row type and all the logic associated to one row.
-There is currently one way to do this if you are using cells that are instantiated from nib files.
+There is currently one way to do this **if you are using cells that are instantiated from nib files**. Currently, none of Eureka's core rows are instantiated from nib files but some of the custom rows in [EurekaCommunity] are, in particular the [PostalAddressRow](https://github.com/EurekaCommunity/PostalAddressRow) which was moved there.
 
-What you have to do is define a subclass and a nibfile with a cell of that subclass type. This will create a row with that cell.
+What you have to do is:
+* Create a nib file containing the cell you want to create.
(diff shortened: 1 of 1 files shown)


mtnbarreto:
👍

PR #739, opened 2016-10-18
Added an empty implementation for `tableView(_:willDisplay:forRowAt:)`

apauly:
With the current version of Swift, it seems that subclasses of `FormViewController` are not able to implement methods from protocols that are originally implemented by `FormViewController`, unless they explicitly override a method from their superclass.

Without a blank implementation, there's (currently) no way to implement `tableView(_:willDisplay:forRowAt:)`. Therefore, I've just added a blank implementation.

In Addition to that, I removed `internal(set)` from the property `validationErrors`. Is there any reason why it is declared internal? I'm currently using it to implement server side validation.

### DIFF
--- a/Source/Core/BaseRow.swift
+++ b/Source/Core/BaseRow.swift
@@ -38,7 +38,7 @@ open class BaseRow : BaseRowType {
     
     public var validationOptions: ValidationOptions = .validatesOnBlur
     // validation state
-    public internal(set) var validationErrors = [ValidationError]() {
+    public var validationErrors = [ValidationError]() {
         didSet {
             guard validationErrors != oldValue else { return }
             RowDefaults.onRowValidationChanged["\(type(of: self))"]?(baseCell, self)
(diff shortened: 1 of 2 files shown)


mtnbarreto:
Hi @apauly , thanks for contributing!!

I would prefer to keep `validationErrors ` internal unless you give me a good reason to change it. It is not intended to be set up programmatically. I would consider to make it public if you provide more details of what you want to accomplish, the code and the trick you are ding to do server side validations using it. 

Regarding the `tableView:willDisplayCell:forRowAtIndexPath` method, you should interoperate with obj-c properly by declaring objc attribute as shown bellow. 

```swift
@objc(tableView:willDisplayCell:forRowAtIndexPath:) func tableView(_ tableView: UITableView, willDisplay cell: UITableViewCell, forRowAt indexPath: IndexPath) {
      print("HI")
}
```

Let me know if it works

Issue #747, opened 2016-10-20
A new rule - Row equals to Row


skrosoft:
Hello,

It could be usefull a validator that allow us to compare 2 row values, example with the field Password and PasswordConfirm, both field must have the same value to make data valid.

`$0.add(rule: RuleEqualsToRow(msg:"Password doesnt match.", tag:"password"))`

Vincent

mtnbarreto:
@skrosoft  Makes total sense, Could you implement it and make a pull request?

skrosoft:
#750 here it is the pull request. @mtnbarreto please let me know if it seems good for you.

PR #750, opened 2016-10-22
RuleEqualsToRow validation feature

skrosoft:
#747 - A new rule validation. Compare that two row has the same values. Maybe we should create also a RuleNotEqualsToRow, for the moment, I dont think this is usefull.

### DIFF
--- a/Source/Validations/RuleEqualsToRow.swift
+++ b/Source/Validations/RuleEqualsToRow.swift
@@ -26,14 +26,14 @@ import Foundation
 
 public struct RuleEqualsToRow<T: Equatable>: RuleType {
     
-    public init(msg: String = "Fields doenst match!", form: Form, tag: String){
+    public init(form: Form, tag: String, msg: String = "Fields doesn't match!"){
         self.validationError = ValidationError(msg: msg)
         self.form = form
         self.tag = tag
         self.row = nil;
     }
     
-    public init(msg: String = "Fields doenst match!", row: BaseRow?){
+    public init(row: RowOf<T>, msg: String = "Fields doesn't match!"){
         self.validationError = ValidationError(msg: msg)
         self.form = nil
         self.tag = nil
@@ -44,7 +44,7 @@ public struct RuleEqualsToRow<T: Equatable>: RuleType {
     public var validationError: ValidationError
     public var form: Form?
     public var tag: String?
-    public var row: BaseRow?
+    public var row: RowOf<T>?
     
     public func isValid(value: T?) -> ValidationError? {
         
@@ -54,7 +54,7 @@ public struct RuleEqualsToRow<T: Equatable>: RuleType {
             _row = form!.rowBy(tag: tag!);
         }
         
-        if (_row!.baseValue as? T == value){
+        if (_row?.value == value){
             return nil;
         }else{
             return validationError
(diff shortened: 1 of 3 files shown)


Source/Validations/RuleEqualsToRow.swift
```diff
@@ -0,0 +1,63 @@
...
+public struct RuleEqualsToRow<T: Equatable>: RuleType {
+    
+    public init(msg: String = "Fields doenst match!", form: Form, tag: String){
+        self.validationError = ValidationError(msg: msg)
+        self.form = form
+        self.tag = tag
+        self.row = nil;
+    }
+    
+    public init(msg: String = "Fields doenst match!", row: BaseRow?){
```
mtnbarreto: should be..

```swift
public init(msg: String = "Fields doesn't match!", row: RowOf<T>)
```

instead of 
```swift
public init(msg: String = "Fields doenst match!", row: BaseRow?){
```

Source/Validations/RuleEqualsToRow.swift
```diff
@@ -0,0 +1,63 @@
...
+    
+    public func isValid(value: T?) -> ValidationError? {
+        
+        var _row = row;
+        
+        if (_row == nil){
+            _row = form!.rowBy(tag: tag!);
+        }
+        
+        if (_row!.baseValue as? T == value){
```
mtnbarreto: `_row?.value` should work..

skrosoft:
Hi @mtnbarreto, here it is, hope its ok! :)

Source/Validations/RuleEqualsToRow.swift
```diff
@@ -0,0 +1,63 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+import Foundation
+
+public struct RuleEqualsToRow<T: Equatable>: RuleType {
+    
+    public init(form: Form, tag: String, msg: String = "Fields doesn't match!"){
```
mtnbarreto: Sorry, actually it's "don't". I will make the change.

Source/Validations/RuleEqualsToRow.swift
```diff
@@ -0,0 +1,63 @@
...
+    public var form: Form?
+    public var tag: String?
+    public var row: RowOf<T>?
+    
+    public func isValid(value: T?) -> ValidationError? {
+        
+        var _row = row;
+        
+        if (_row == nil){
+            _row = form!.rowBy(tag: tag!);
```
mtnbarreto: There is no need to use semicolon in swift.

yihyang:
Hi, May I know what is the status of this pull request? I have loaded version 2.0.0-beta.1 but this file is not included inside the podfile, thanks.

kwstasna:
@yihyang in my version either. I have to put it by myself.

mtnbarreto:
what about 2.0.1 version?

Issue #783, opened 2016-11-10
NavigationController interactive pop gesture could cause navbar blink and toolbar animation interruption


banjun:
Environment:
Eureka 2.0.0-beta.1, Xcode 8.1, iOS 9 and 10

Reproducible View Controller Hierarchy:
contained in UINavigationController that have viewControllers:
`... → Eureka.FormViewController → any UIViewController`

from the last view controller, initiating interactive pop gesture and cancelling the gesture could cause the labels on navigation bar and toolbar to blink (disappear and appear animation), and/or, toolbar hidden even if the last view controller set its navigationController.setToolbarHidden to false.
this behaviour seems to be caused by navigation animation interruption, that is caused by the following:

https://github.com/xmartlabs/Eureka/blob/master/Source/Core/Core.swift#L501

```swift
coordinator.animateAlongsideTransition(in: parent?.view, animation: deselectionAnimation, completion: reselection)
```

this was introduced in #406.

I have found a workaround and similar discussion is found at a blog entry (posted by @ZevEisenberg ?)

* workaround: use `animate(alongsideTransition:completion:)` instead of `animateAlongsideTransition(in:animation:completion:)`
* the related blog (comments) https://www.raizlabs.com/dev/2016/05/smarter-animated-row-deselection-ios/

may Eureka use the workaround?

ZevEisenberg:
@banjun sorry, I think I caused this problem. I'm the one who added `animateAlongsideTransition(in:` to Eureka, and I never went back and fixed it after Oleksii posted their comment about `animateAlongsideTransition(`. Eureka should definitely use this workaround! I don't have time to write a PR, but it should be pretty easy to do.

PR #786, opened 2016-11-11
Fix UINavigationController interactive pop gesture could cause navbar blink and toolbar animation interruption

banjun:
use workaround issued in #783.


mtnbarreto:
@banjun  👍 thanks for contributing!

PR #813, opened 2016-11-30
Fix bug: if inserted text that has yet to be confirmed by the user update cell will cause lost typing info.

KevinGong2013:
Fix bug: if inserted text that has yet to be confirmed by the user update cell will cause lost typing info.
bug


mtnbarreto:
@KevinGong2013 Could you explain what is wrong?

KevinGong2013:
This issue has been fixed. The issue caused by update cell, your latest commit removed line: 61 in Row.swift. so we can drop this pr.

Issue #808, opened 2016-11-28
Picker Keyboard


SpencerC:
Hi,
The inline picker controls are cool, but I was wondering if there was an easy way to display the picker in the keyboard area with the prev/next controls like Safari does:
![img_3684](https://cloud.githubusercontent.com/assets/817672/20655831/3bc197f2-b4db-11e6-907b-2c7d5bb7a611.PNG)

If there isn't, and I were to implement an approach, would it be a good candidate for a pull request?

mtnbarreto:
We don't have such picker view row variation yet. We would love to have it so Feel free to make a pull request. 
Maybe PickerInputRow is a good candidate name for the row. What do you think?

PR #818, opened 2016-12-02
Added feature PickerInputRow

SpencerC:
Followed pattern of the generic picker input

#808


SpencerC:
Also added a variable to control the spacing between the keyboard and the selected field when scrolling is required, and added smoothly animated scrolling. Tried to make it another pull request, but github desktop is misbehaving. LMK if that's a problem, and I'll try again

Eureka.xcworkspace/xcshareddata/Eureka.xcscmblueprint
```diff
@@ -0,0 +1,30 @@
+{
```
mtnbarreto: This file should be added to gitignore.

Source/Core/Core.swift
```diff
@@ -840,7 +846,7 @@ extension FormViewController {
             table.contentInset = tableInsets
             table.scrollIndicatorInsets = scrollIndicatorInsets
             if let selectedRow = table.indexPath(for: cell) {
-                table.scrollToRow(at: selectedRow, at: .none, animated: false)
+                table.scrollToRow(at: selectedRow, at: .none, animated: animateScroll)
```
mtnbarreto: I'm not sure of passing true as animated value works in all cases. In particular when we navigate to a row that is not being displayed in the screen. Have you tested it?

Source/Core/Core.swift
```diff
@@ -431,6 +431,12 @@ open class FormViewController : UIViewController, FormViewControllerProtocol {
         }
     }
     
+    /// Extra space to leave between between the row in focus and the keyboard
+    open var rowKeyboardSpacing: CGFloat = 0
```
mtnbarreto: 👍 Please add associated documentation in readme.

Source/Core/Core.swift
mtnbarreto: i'm fine with this if it works. Please document it in readme.

Source/Rows/PickerInputRow.swift
```diff
@@ -0,0 +1,132 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+import Foundation
+
+class PickerInputTextField: UITextField {
```
mtnbarreto: We don't need to add a UITextField to present the picker control. Take a look at `DateCell` that also has a picker and shows it by implementing these 2 methods....

```swift
 open override func cellCanBecomeFirstResponder() -> Bool {
        return canBecomeFirstResponder
    }

    override open var canBecomeFirstResponder: Bool {
        get {
            return !row.isDisabled
        }
    }
```

By doing so we can use the standard UIKit table view cell.

Source/Rows/PickerInputRow.swift
```diff
@@ -0,0 +1,132 @@
...
+        textField.clearButtonMode = .never
+        if row.isHighlighted {
+            textFieldStartColor = textField.textColor
+            textField.textColor = textField.tintColor
+        } else if textFieldStartColor != nil {
+            textField.textColor = textFieldStartColor
+        }
+        
+        textLabel?.text = nil
+        detailTextLabel?.text = nil
```
mtnbarreto: if row.title is set up we should show the value using `detailTextLabel` otherwise use the `textLabel`.

SpencerC:
@mtnbarreto All should be updated per your feedback. Let me know how it looks!

mtnbarreto:
Hi @SpencerC! Now it looks fine! Great work!!!

PR #821, opened 2016-12-02
Fix issue when showing StepperRow value. Now we use displayValueFor a…

mtnbarreto:
…s we do with the rest of rows.


mtnbarreto:
fix #821

PR #860, opened 2016-12-31
Implement init?(coder: NSCoder) for some cell types

dawiddr:
**The problem**
I've created a custom cell by:
- implementing a cell class that inherits `ButtonCellOf`
- implementing a row class that inherits from `Eureka._ButtonRowOf`

The row class instantiates `CellProvider` and passes a NIB name.

Now, this used to work fine in Eureka 1.7. However, after I updated to the latest Swift 3 master, the following problem has arisen:
- my custom cell implementation wouldn't compile unless `init?(coder: NSCoder)` was implemented
- when I've implemented it by calling `init?(coder: NSCoder)` on `super`, the app has started to crash.

**The solution**
I think that the problem exists because `init?(coder: NSCoder)` in `ButtonCellOf` calls `fatalError` instead of calling `init` on `super`. I suspect that it has been introduced while migrating to Swift 3. This pull request changes that behaviour. 

I've applied the same change to other cell classes that have trivial initialisers.

### DIFF
--- a/Source/Rows/ActionSheetRow.swift
+++ b/Source/Rows/ActionSheetRow.swift
@@ -32,7 +32,7 @@ open class AlertSelectorCell<T: Equatable> : Cell<T>, CellType {
     }
     
     required public init?(coder aDecoder: NSCoder) {
-        fatalError("init(coder:) has not been implemented")
+        super.init(coder: aDecoder)
     }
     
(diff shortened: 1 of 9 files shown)


mtnbarreto:
@dawiddr thanks for contributing!

PR #873, opened 2017-01-10
Fixed the documentation to get the example to work

alper:


alper:
The fail is random, but my commit is a one-line documentation patch anyway.

mtnbarreto:
We have also to change a little bit the form property assignment function to make it work so we avoid this  random failure. @mats-claassen Could be related with the issue you showed me earlier today?

PR #884, opened 2017-01-16
Fix a crash due to uneven KVO observer calls

pnc:
If the application wasn't active when the `FormViewController` was displayed (such as if a system permission dialog was currently active), `FieldRow` would observe `titleLabel.text` twice (once in setup and once in the `NSNotification` callback). `FieldRow` would only call `removeObserver` once in `deinit`, which lead to a crash when the underlying label was deallocated and KVO noticed the dangling observation.

Happy to try to reproduce this in a regression test case if you like.


kuquert:
We teste here, simulated some system dialogs and AppDelegate view calls, it works fine. Thanks for that.

txaiwieser:
I was having the same issue, crashing while going on multitasking and others dialogs, this PR solved my problem

PR #895, opened 2017-01-24
Changed the Access Control of Class RuleRegExp to be open so it can b…

salanswift:
I think the RuleRegExp should be Open so we can make its extension within our project without making any change in the Eureka.


mtnbarreto:
makes sense! thanks!

Issue #261, opened 2016-02-12
PushRow - Prevent Nil


xom-pschiess:
This may be a basic question - but for a PushRow - is there a way to make it to where if I select the same row twice, it doesn't "unselect" the row?

Basically I don't want them to be able to unselect values from PushRows... more similar to how iOS does it on forms such as "New Event" on Calendar.

mats-claassen:
Currently the PushRow is using SelectorViewController which uses a `SelectableSection` which supports enabling or disabling this feature. But there is no way to change the default for now. We might add it in the future.

stefanhong:
+1
By default PushRow should not allow nil selection.  Single selection usually means exactly one option needs to be selected.

mats-claassen:
Merged to Swift3 branch in #551. Will be supported as soon as Swift 3 becomes official

Laptopmini:
For Swift 2.2, you can achieve the same affect by using:
```
.onChange({ row in
    if row.value == nil {
        row.value = // set to previously used value or default
        row.reload()
    }
}
```

andrewprocter:
Thanks @Laptopmini, you're missing a `)` at the end there though 🤓

Laptopmini:
Fixed. Good catch @andrewprocter

mohan-shyam:
@mtnbarreto , Is there a similar implementation for `MultipleSelectorRow`, like you've mentioned earlier in https://github.com/xmartlabs/Eureka/issues/261#issuecomment-255439929 ?

I'd want to make sure that at least one is selected or having a default 'None' or 'Never' option if deselection is allowed?

mohan-shyam:
@mats-claassen and @mtnbarreto , I've noticed in `Eureka/Source/Core/SelectableSection.swift`, the below code.

```
public enum SelectionType {
    
    /**
     * Multiple options can be selected at once
     */
    case multipleSelection
    
    /**
     * Only one selection at a time. Can additionally specify if deselection is enabled or not.
     */
    case singleSelection(enableDeselection: Bool)
}
```

Which means, `enableDeselection` hasn't been implemented yet for `multipleSelection`!! Is there any chance that it would be?

Thank you both, for the amazing work on Eureka.

mats-claassen:
@mohan-shyam 
I think normally in a multiple selection you should allow to deselect all options. You can later validate that at least one is selected and point the user to that row if he did not select any option.

jaimeagudo:
@mtnbarreto 
[...]

PR #908, opened 2017-01-30
Always call onSelectSelectableRow, added properties to control selector vc dismissal

ilyapuchka:
This PR reverts some of the changes made in #551 and adds two new properties for better control of dismissing `SelectorViewController`. Please see [this](https://github.com/xmartlabs/Eureka/pull/551#issuecomment-262821225) and [this](https://github.com/xmartlabs/Eureka/issues/261#issuecomment-274661489) comments for some reasoning.

### DIFF
--- a/Example/Example/ViewController.swift
+++ b/Example/Example/ViewController.swift
@@ -665,6 +665,7 @@ class NativeEventFormViewController : FormViewController {
                     $0.value = .Never
                 }.onPresent({ (_, vc) in
                     vc.enableDeselection = false
+                    vc.dismissOnSelection = false
                 })
         
         form +++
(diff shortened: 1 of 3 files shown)


mtnbarreto:
@ilyapuchka great work!

Are these properties default values the right ones to ensure backwards compatibility?

```swift
public var dismissOnSelection = true
public var dismissOnChange = true
```

I want to make sure we are not changing default behaviour to current eureka users.

ilyapuchka:
To preserve current behaviour `dismissOnSelection` should be false, but I would say that `true` is a better default. Don't have problems with changing it anyway.

Issue #80, opened 2015-11-11
how to change Section height


lkqin:
how to change Section height

mats-claassen:
A section consists of a header, rows and a footer. Each cell has a height variable so you can set the height of each row.
If you are setting the header and footer with strings you can also set the height for that header or footer in the initializer like this:
```
Section("FieldRow examples") {
        $0.header?.height = { 100 }
}
```
You can also set the header or footer like this if you have a custom view:
```
var header = HeaderFooterView<MyHeaderNibFile>(.NibFile(name: "MyHeaderNibFile", bundle: nil))        
header.onSetupView = { view, _, _ in
     // customize header
 }

header.height = { 100 }
section.header = header
```

iBenjamin:
if you want to make a section header height to 0, you can use to following code.

```objc
$0.header = HeaderFooterView<UIView>(HeaderFooterProvider.Class)
$0.header?.height = { CGFloat.min }
```

PR #912, opened 2017-01-31
Do not leave a space for section header/footer if they don't have a title

ilyapuchka:


mtnbarreto:
Often we use sections just to group rows, these rows don't have any title. 

Easier way to set up a 0 height section is https://github.com/xmartlabs/Eureka/issues/80#issuecomment-226708469

Eventually if we have many sections with 0 height in our app, we can either extend Section type or add a new convenience initializer.

Any thoughts?

ilyapuchka:
It might make sense to have empty string as a default parameter for header title.
I just noticed that though I don't want any footer and I don't explicitly set it up in any way I would not expect extra space.
I would say that the behaviour in case there is no title for header/footer should be the same as the behaviour of standard table view controller that has nil title for section (as `titleForHeaderInSection` returns `String?`), not an empty string.

ilyapuchka:
The default behaviour of table view with `nil` or empty footer/header still creates a space, so this change is not needed.

PR #917, opened 2017-02-03
Ability to use InterfaceBuilder to specify almost every row design.

mtnbarreto:


Source/Rows/PickerInputRow.swift
```diff
@@ -28,19 +28,20 @@ import Foundation
 
 open class PickerInputCell<T: Equatable> : Cell<T>, CellType, UIPickerViewDataSource, UIPickerViewDelegate where T: Equatable, T: InputTypeInitiable {
     
-    public var picker: UIPickerView
+    lazy public var picker: UIPickerView = {
```
mats-claassen: I remember that we had issues with lazy variables using Cocoapods. Did you test that this works correctly?

Source/Rows/StepperRow.swift
```diff
@@ -12,42 +12,41 @@ import UIKit
...
+        self.stepper = stepper
         self.stepper.translatesAutoresizingMaskIntoConstraints = false
-
-        self.valueLabel = UILabel()
-        self.valueLabel.translatesAutoresizingMaskIntoConstraints = false
-        self.valueLabel.numberOfLines = 1
-
+        
+        let valueLabel = UILabel()
+        self.valueLabel = valueLabel
```
mats-claassen: Why don't you assign this directly?

Source/Rows/SwitchRow.swift
```diff
@@ -28,26 +28,24 @@ import Foundation
 
 open class SwitchCell : Cell<Bool>, CellType {
     
-    public typealias Value = Bool
+    @IBOutlet public weak var switchControl: UISwitch!
     
     required public init(style: UITableViewCellStyle, reuseIdentifier: String?) {
         super.init(style: style, reuseIdentifier: reuseIdentifier)
+        let switchC = UISwitch()
+        switchControl = switchC
```
mats-claassen: Also here, why not `switchControl = UISwitch()`?

PR #965, opened 2017-03-17
Suggestion: FieldRow.titlePercentage instead of FieldRow.textFieldPercentage

quiekedei:
### Problem:
When using `FieldRow` with a left-aligned `textField`, the layout (in particular the `textField`'s x-position) changes when adding an accessory view or an accessory type. This is due to the fact that `textFieldPercentage` is calculated based on the `contentView`'s width which shrinks when an accessory is in place.
Now when using multiple `FieldRow`s, the horizontal alignment of the `textField`s breaks as soon as one of them shows an accessory (e.g. an error indicator) — an unfavorable behavior.

### Solution:
I propose a solution which drops `textFieldPercentage` in favor of `titlePercentage`, i.e. defining the title's width as a percentage of the cell's overall width. Note that the title encompasses both the `titleLabel` as well as an optional `imageView`. This way the relative position of the `textField` within the cell remains fixed.

### Demo:
I created a quick Demo illustrating the issue as well as the solution (see attachment). Please update the `Podfile` and run `pod install` to switch between the latest **Eureka** release and the version containing the suggested fix.

[EurekaFieldRowIssueDemo.zip](https://github.com/xmartlabs/Eureka/files/849981/EurekaFieldRowIssueDemo.zip)

### Caveats:
When using a right-to-left interface layout (e.g. Arabic) this fix only works for iOS versions 9.0 or higher. This is due to the fact that **Eureka** allows only app extension API, which makes it impossible to determine interface layout direction on lower iOS versions.


Source/Rows/Common/FieldRow.swift
```diff
@@ -381,6 +391,30 @@ open class _FieldCell<T> : Cell<T>, UITextFieldDelegate, TextFieldCell where T:
...
+		super.layoutSubviews()
+		guard let row = (row as? FieldRowConformance) else { return }
+		guard let titlePercentage = row.titlePercentage else {
+			calculatedTitlePercentage = nil
+			return
+		}
+		var targetTitleWidth = bounds.size.width * titlePercentage
+		if let imageView = imageView, let _ = imageView.image, let titleLabel = titleLabel {
+			var extraWidthToSubtract = titleLabel.frame.minX - imageView.frame.minX // Left-to-right interface layout
+			if #available(iOS 9.0, *) {
```
mtnbarreto: Why don't use `UIApplication.shared.userInterfaceLayoutDirection == .rightToLeft` to determine layout direction in iOS 8 ?

Source/Rows/Common/FieldRow.swift
```diff
@@ -268,28 +287,19 @@ open class _FieldCell<T> : Cell<T>, UITextFieldDelegate, TextFieldCell where T:
...
                 dynamicConstraints += NSLayoutConstraint.constraints(withVisualFormat: "H:[imageView]-(15)-[label]-[textField]-|", options: NSLayoutFormatOptions(), metrics: nil, views: views)
-                dynamicConstraints.append(NSLayoutConstraint(item: textField,
-                                                             attribute: .width,
-                                                             relatedBy: (row as? FieldRowConformance)?.textFieldPercentage != nil ? .equal : .greaterThanOrEqual,
-                                                             toItem: contentView,
-                                                             attribute: .width,
-                                                             multiplier: (row as? FieldRowConformance)?.textFieldPercentage ?? 0.3,
-                                                             constant: 0.0))
-            } else {
+                dynamicConstraints.append(NSLayoutConstraint(item: titleLabel, attribute: .width, relatedBy: calculatedTitlePercentage != nil ? .equal : .greaterThanOrEqual, toItem: contentView, attribute: .width, multiplier: calculatedTitlePercentage ?? 0.3, constant: 0.0))
```
mtnbarreto: shouldn't be 'calculatedTitlePercentage ?? 0.7' instead of `calculatedTitlePercentage ?? 0.3`?

quiekedei: Since this constraint is applied to `titleLabel` instead of `textField`, the inverse value (i.e. 0.3) should be correct IMHO.
However, your comment did point me to a flaw in my solution when it comes to handling nil values. See [this commit](https://github.com/xmartlabs/Eureka/pull/965/commits/0a8c1a1a52df3fa204c63702f49bf340357edcb9) for more details.

quiekedei: See [previous comment](https://github.com/xmartlabs/Eureka/pull/965#discussion_r129263093)

Source/Rows/Common/FieldRow.swift
quiekedei: As pointed out in the PR description, **Eureka** only allows app extension API, which means the solution you proposed is not possible.
We could fiddle around with `NSLocale` and device languages but I would not consider any of this a truly stable solution.

If you know of any other way to determine layout direction, please let me know about it. I'd be happy to include it.

mtnbarreto:
Sometimes we just want to stick the label to the text field without any a alignment. To do so we need to set `titlePercentage` to nil so it can be non-optional neither set up the default value in layout subviews.  
I'm gonna fix the code so we can merge out this.

Source/Rows/Common/FieldRow.swift
```diff
@@ -86,8 +86,25 @@ open class FieldRow<Cell: CellType>: FormatteableRow<Cell>, FieldRowConformance,
     open var keyboardReturnType: KeyboardReturnTypeConfiguration?
 
     /// The percentage of the cell that should be occupied by the textField
-    open var textFieldPercentage: CGFloat?
-
+	@available (*, deprecated, message: "Use titleLabelPercentage instead")
+	open var textFieldPercentage : CGFloat? {
+		get {
+			guard let titlePercentage = titlePercentage else { return nil }
+			return 1 - titlePercentage
```
mtnbarreto: `return titlePercentage.map { 1 - $0}` looks better

Source/Rows/Common/FieldRow.swift
```diff
@@ -86,8 +86,25 @@ open class FieldRow<Cell: CellType>: FormatteableRow<Cell>, FieldRowConformance,
...
+		}
+		set {
+			if let newValue = newValue {
+				titlePercentage = 1 - newValue
+			}
+			else {
+				titlePercentage = nil
+			}
```
mtnbarreto: `titlePercentage = newValue.map { 1 - $0 }`

Source/Rows/Common/FieldRow.swift
```diff
@@ -269,28 +288,23 @@ open class _FieldCell<T> : Cell<T>, UITextFieldDelegate, TextFieldCell where T:
...
-                                                             attribute: .width,
-                                                             relatedBy: (row as? FieldRowConformance)?.textFieldPercentage != nil ? .equal : .greaterThanOrEqual,
-                                                             toItem: contentView,
-                                                             attribute: .width,
-                                                             multiplier: (row as? FieldRowConformance)?.textFieldPercentage ?? 0.3,
-                                                             constant: 0.0))
-            } else {
+				if let calculatedTitlePercentage = calculatedTitlePercentage {
+					dynamicConstraints.append(NSLayoutConstraint(item: titleLabel, attribute: .width, relatedBy: .equal, toItem: contentView, attribute: .width, multiplier: calculatedTitlePercentage, constant: 0.0))
+				}
```
mtnbarreto:
```swift
if let calculatedTitlePercentage = calculatedTitlePercentage {
					dynamicConstraints.append(NSLayoutConstraint(item: titleLabel, attribute: .width, relatedBy: .equal, toItem: contentView, attribute: .width, multiplier: calculatedTitlePercentage, constant: 0.0))
}
```
should be replaced by

```swift
dynamicConstraints.append(NSLayoutConstraint(item: titleLabel, attribute: .width, relatedBy: (calculatedTitlePercentage != nil ? NSLayoutRelation.equal : NSLayoutRelation.lessThanOrEqual), toItem: contentView, attribute: .width, multiplier: calculatedTitlePercentage ?? 0.7, constant: 0.0))
```

Source/Rows/Common/FieldRow.swift
```diff
@@ -269,28 +288,23 @@ open class _FieldCell<T> : Cell<T>, UITextFieldDelegate, TextFieldCell where T:
...
                 dynamicConstraints += NSLayoutConstraint.constraints(withVisualFormat: "H:|-[label]-[textField]-|", options: [], metrics: nil, views: views)
-                dynamicConstraints.append(NSLayoutConstraint(item: textField,
-                                                             attribute: .width,
-                                                             relatedBy: (row as? FieldRowConformance)?.textFieldPercentage != nil ? .equal : .greaterThanOrEqual,
-                                                             toItem: contentView,
-                                                             attribute: .width,
-                                                             multiplier: (row as? FieldRowConformance)?.textFieldPercentage ?? 0.3,
-                                                             constant: 0.0))
-            } else {
+				if let calculatedTitlePercentage = calculatedTitlePercentage {
```
mtnbarreto: same above comment here

Source/Rows/Common/FieldRow.swift
```diff
@@ -382,6 +396,27 @@ open class _FieldCell<T> : Cell<T>, UITextFieldDelegate, TextFieldCell where T:
...
     open func textFieldShouldEndEditing(_ textField: UITextField) -> Bool {
         return formViewController()?.textInputShouldEndEditing(textField, cell: self) ?? true
-    }
-
+	}
+	
+	open override func layoutSubviews() {
+		super.layoutSubviews()
+		guard let row = (row as? FieldRowConformance) else { return }
+		let titlePercentage = row.titlePercentage ?? 0.3
```
mtnbarreto: should be replaced by 
```swift
guard let titlePercentage = row.titlePercentage else  { return }
```

quiekedei:
I get what you mean with not having alignment among the fields at all and I just realized that I apparently had the default relation between `titleLabel` and `textField` wrong, i.e. if `titlePercentage` is not set, `titleLabel` should occupy 70% of the space or less. I made appropriate adjustments in the code.
However, in `customConstraints()` you are already using a default value right now — `…multiplier: 0.3…` with `relatedBy: .greaterThanOrEqual` if `textFieldPercentage` *is not* set. I argue that it would be better to use the `calculatedTitlePercentage` instead of the hard-coded constant and move the constant to `layoutSubviews`. This way the behavior for `titlePercentage = nil` does not change but we still get the benefit of the fix I suggested.
I made the respective suggestions in the code comments. Let me know what you think.

Source/Rows/Common/FieldRow.swift
quiekedei: As described [here](https://github.com/xmartlabs/Eureka/pull/965#issuecomment-318293995), I would go for this

```
if let calculatedTitlePercentage = calculatedTitlePercentage {
	dynamicConstraints.append(NSLayoutConstraint(item: titleLabel,
	                                             attribute: .width,
	                                             relatedBy: (row as? FieldRowConformance)?.titlePercentage != nil ? .equal : .lessThanOrEqual,
	                                             toItem: contentView,
	                                             attribute: .width,
	                                             multiplier: calculatedTitlePercentage,
	                                             constant: 0.0))
}
```

Source/Rows/Common/FieldRow.swift
quiekedei: See [this comment](https://github.com/xmartlabs/Eureka/pull/965#discussion_r129781083)

Source/Rows/Common/FieldRow.swift
quiekedei: As described [here](https://github.com/xmartlabs/Eureka/pull/965#issuecomment-318293995), I would go for this:

`let titlePercentage = row.titlePercentage ?? 0.7`

Source/Rows/Common/FieldRow.swift
quiekedei: Given that `titlePercentage`  is a `CGFloat`, I do not see how we could use `map` on it. Am I missing something here?

If you would like to opt for a more concise style, I could offer the following one-liner:

```
return titlePercentage != nil ? 1 - titlePercentage! : nil
```

Source/Rows/Common/FieldRow.swift
quiekedei: Similar to [this comment](https://github.com/xmartlabs/Eureka/pull/965#discussion_r129782699).

```
titlePercentage = newValue != nil ? 1 - newValue! : nil
```

Source/Rows/Common/FieldRow.swift
mtnbarreto: Look at https://www.natashatherobot.com/swift-using-map-to-deal-with-optionals/ that shows how to use map with optionals.

quiekedei: Oh nice! Didn't know about, thanks for the hint! 👍

quiekedei:
Sorry! Accidentally clicked the wrong button while commenting. Didn't mean to close this PR.

Source/Rows/Common/FieldRow.swift
mtnbarreto: So let's make `calculatedTitlePercentage` non optional and set up 0.7 as default value. Not sure of the code works for all cases.

PR #985, opened 2017-04-06
Overload <<< operator to append multiple rows

lightsprint09:
Instead of using the following:
```swift
form +++ SelectableSection<ListCheckRow<String>>("Where do you live", selectionType: .singleSelection(enableDeselection: true))

let continents = ["Africa", "Antarctica", "Asia", "Australia", "Europe", "North America", "South America"]
for option in continents {
    form.last! <<< ListCheckRow<String>(option){ listRow in
        listRow.title = option
        listRow.selectableValue = option
        listRow.value = nil
    }
}
```

on could do the following as well, which can be used without a reference to form and a force unwrapped reference to the last section:
```swift
form +++ SelectableSection<ListCheckRow<String>>("Where do you live", selectionType: .singleSelection(enableDeselection: true))

<<<  ["Africa", "Antarctica", "Asia", "Australia", "Europe", "North America", "South America"].map { title in 
   return ListCheckRow<String>(title){ listRow in
        listRow.title = title
        listRow.selectableValue = title
        listRow.value = nil
    }
}
```

### DIFF
--- a/Source/Core/Operators.swift
+++ b/Source/Core/Operators.swift
@@ -141,6 +141,21 @@ public func <<< (left: BaseRow, right: BaseRow) -> Section {
 }
 
 /**
+ Appends a collection of rows to a section.
+ 
+ - parameter left:  the section
+ - parameter right: the collection of rows to be appended
+ 
+ - returns: the section
+ */
+@discardableResult
+public func <<< <C: Collection>(left: Section, right: C) -> Section where C.Iterator.Element == BaseRow {
+    left.append(contentsOf: right)
+    
+    return left
+}
+
+/**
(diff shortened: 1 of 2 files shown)


mtnbarreto:
Hi @lightsprint09!

I like the idea of being able to add collections of rows to section but is something very straightforward to add from your own application code.

I would prefer to expose only the essential/basic operator that we need to create a form. Anyone is able to implement operator variants if desired.

Cheers!

lightsprint09:
Hey thanks for your feedback.

We think this is a important operator. Because is the equivalent to the already standard method `append(contentsOf:)` which everyone expects on a collection.

We see a form as a collection :)

Issue #989, opened 2017-04-07
header.onSetupView called multiple times


thierrybucco:
Hi,

In my project I am creating a custom section from a xib.

```swift
import UIKit
import Eureka

class ViewController: FormViewController {

	override func viewDidLoad() {
		super.viewDidLoad()
		addSection1()
	}

	private func addSection1() {
		let informationsSection = makeSectionView(title: "SECTION1")
		self.form +++ informationsSection
	}

	private func makeSectionView(title: String) -> Section {

		return Section() { section in
			var header = HeaderFooterView<CustomSectionView>(.nibFile(name: "CustomSectionView", bundle: nil))

			header.onSetupView = { view, _ in
				view.backgroundColor = UIColor.groupTableViewBackground
				view.titleLabel.text = title
				print("onSetupView")
			}
			section.header = header
		}
	}
}
```

I discovered that onSetupView is called each time view is being displayed .
For what I understand, onSetupView should be called once at creation time.

/// Closure called when the view is created. Useful to customize its appearance.

Do you have an idea?

Here is link to test project: https://www.dropbox.com/s/oiq1sn87l8r2gsq/onSetupViewTest.zip?dl=0

Thanks!

thierrybucco:
After investigation it seems that onSetupView should be called after createView and not at the end.
What do you think?

```
public func viewForSection(_ section: Section, type: HeaderFooterType) -> UIView? {
        var view: ViewType?
        if type == .header {
            view = section.headerView as? ViewType ?? {
                            let result = viewProvider?.createView()
                            section.headerView = result
				onSetupView?(result!, section)
                            return result
                        }()
        }
        else {
            view = section.footerView as? ViewType ?? {
                            let result = viewProvider?.createView()
                            section.footerView = result
				onSetupView?(result!, section)
                            return result
                        }()
        }
        guard let v = view else { return nil }
        //onSetupView?(v, section)
        return v
    }
```

thierrybucco:
@mats-claassen could you investigate this ticket and maybe accept my pull request? 
Thanks a lot.

santiagofm:
Hey @thierrybucco ! This seems like an issue to me. Would you mind updating the PR ?
Cheers

thierrybucco:
Hello @santiagofm I updated PR I hope all is ok.
Have a nice day.

m-revetria:
I'm closing this as it is by design.

thierrybucco:
[...]

PR #999, opened 2017-04-11
Fixed onSetupView called multiple times and not only on view creation…

thierrybucco:
…. #989


santiagofm:
hey @thierrybucco, would you like to update this PR so we can review it ?

mtnbarreto:
Hi @thierrybucco , As Eureka documentation indicates, `onSetupView` is called each time header/footer view appears on screen.

```swift
Section() { section in
    var header = HeaderFooterView<MyHeaderNibFile>(.nibFile(name: "MyHeaderNibFile", bundle: nil))

    // Will be called every time the header appears on screen
    header.onSetupView = { view, _ in
        // Commonly used to setup texts inside the view
        // Don't change the view hierarchy or size here!
    }
    section.header = header
}
```

diegogarciar:
Hello, I'm having the same issue again, I reviewed the code and it is the same as it was before this commit 
https://github.com/xmartlabs/Eureka/pull/999/commits/0289b9ba26d328f219b290eb62e2161f5c67e3fc
Just wanted to check if this fix/commit was wrong somehow or if it there was a wrong merge somewhere

Issue #1000, opened 2017-04-12
memory leaks 


alexsanderkhitev:
Hello! I have a memory leak in Eureka framework. I always use weak links everywhere, while I do not use Form anywhere in the blocks. How can I fix it?

![memory xcworkspace 2017-04-12 16-41-09](https://cloud.githubusercontent.com/assets/12511909/24958074/1c71ef76-1f9f-11e7-81ee-d9cc0598a19c.png)

mtnbarreto:
Hi @alexsanderkhitev I will look into it when i have a chance. 
Could you please share the Eureka form that is causing the memory leak?? Which is the eureka version ?

mtnbarreto:
same as #974

alexsanderkhitev:
@mtnbarreto I can share my code 

I use Eureka (2.0.1)

```
class EventsUpcomingOwnerVC: FormViewController {
    
    fileprivate enum SectionType: String {
        case main, requests, approved, navigation, settings
    }
    
    // MARK: - Data
    
    var card: CardModel!
    
    // MARK: - Delegates
    
    weak var delegate: EventsUpcomingOwnerVCDelegate?
    
    // MARK: - Lifecycle

    override func viewDidLoad() {
        super.viewDidLoad()
        setupSettings()
        // ui
        setupUISettings()
        setupNavSettings()
        setupForm()
        // observers
//        firObserve(true)
        debugPrint("FormViewController init")
    }
    
    deinit {
        guard card != nil else { return }
        FIRCardDatabaseManager.shared.observeInterestedRequests(card, isObserve: false, getNewRequest: nil)
        debugPrint("FormViewController deinit")
    }

    private func setupSettings() {
        definesPresentationContext = true
    }
    
    // MARK: - UI
    
    private func setupUISettings() {
        view.backgroundColor = .white
    }
    
    private func setupNavSettings() {
        guard card != nil else { return }
        navigationItem.title = card.title
    }
    
    // MARK: - Form 
    
    private func setupForm() {
        form
            +++ section(.main, headerHeight: nil, footerHeight: nil)
            <<< cardPlacePhotoRow()
            <<< cardInfoRow()
            <<< cardDetailLocationRow()
        setupRequestsSections()
        setupApprovedSection()
        
        let navigationSection = section(.navigation, headerHeight: 30, footerHeight: nil)
        form
            +++ navigationSection
            <<< upcomingChatRow()
        
[...]

Issue #974, opened 2017-03-25
memory leaks when I call `self` in `onChange`


abellee:
Eureka: 2.0.1
Xcode: 8.2.1
iOS: 8.1

2017-03-25 19:32:23.952 xx[23700:40182275] Memory Leak: (
    "xx.VerifyCodeViewController",
    UIView,
    UITableView,
    UITableViewWrapperView,
    "Eureka.TextCell"
)
2017-03-25 19:32:23.953 xx[23700:40182275] Possibly Memory Leak.
In case that Eureka.TextCell should not be dealloced, override -willDealloc in Eureka.TextCell by returning NO.
View-ViewController stack: (
    "xx.VerifyCodeViewController",
    UIView,
    UITableView,
    UITableViewWrapperView,
    "Eureka.TextCell"
)

I tried to comment all codes maybe cause leaking, but still show me the memory leak after I input something into the `TextCell` and pop my `ViewController`.

Laptopmini:
Make sure you use a weak reference to self to avoid memory leaks.

abellee:
@Laptopmini ya I tried to add everywhere that possible to make leaking but still not work

abellee:
@Laptopmini even though I have delete all my codes just leave the cells

Laptopmini:
I'll jump the gun for the maintainers of this library and recommend you create an example project which reproduces this memory leak and share it here.

Bashta:
Having the same issue! Even after using `[unowned self]` i do get leaks.

alexsanderkhitev:
Hi @Bashta  Have you solved this problem? I also ran into the problem of memory leak.

Bashta:
@alexsanderkhitev No. Even stripping the project down to the simplest demo produces a leak. dropped Eureka altogether since i dont have enough time to investigate where the leak occurs.

alexsanderkhitev:
@Bashta thanks for response

m-revetria:
Hi @abellee, thanks for letting us know about the issue.

We are going deeper into it in order to solve any memory-related issue. I'd like to know how are you getting those leaks log.

mats-claassen:
@abellee I think #1000 and #1001 fix this issue.

PR #1001, opened 2017-04-13
Fixes memory issues

m-revetria:
Improvements done:
* Removed retain cycles in example view controllers
* Removed retain cycles in `RuleEqualsToRow` rule
* Cleaned up collections from KVOWrappers

This might solve #974 and #1000

### DIFF
--- a/Example/Example/ViewController.swift
+++ b/Example/Example/ViewController.swift
@@ -519,9 +519,9 @@ class NavigationAccessoryController : FormViewController {
         
         form = Section(header: "Settings", footer: "These settings change how the navigation accessory view behaves")
             
-             <<< SwitchRow("set_none") {
+             <<< SwitchRow("set_none") { [weak self] in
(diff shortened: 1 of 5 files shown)


Example/Example/ViewController.swift
```diff
@@ -657,8 +657,8 @@ class NativeEventFormViewController : FormViewController {
                         endRow.updateCell()
                     }
                 }
-                .onExpandInlineRow { cell, row, inlineRow in
-                    inlineRow.cellUpdate { [weak self] cell, dateRow in
+                .onExpandInlineRow { [weak self] cell, row, inlineRow in
+                    inlineRow.cellUpdate() { cell, row in
```
mtnbarreto: we don't need the `()` after cellUpdate method. Anyway let's keep this code anyway.

mtnbarreto:
This pull request should fix some memory leaks reported in #1000 and #974.

PR #1002, opened 2017-04-14
Updated README.md to fix install issue with cocoapods

chrishorton:
Making it easier for future users to install


mtnbarreto:
Hi @chrishorton I would prefer to remove the pod version so cocoapods installs the latest official release which is a stable eureka version. What do you think?

mtnbarreto:
Developer should be in charge of setting and fixing the pod version.

chrishorton:
I agree, that would be a better, more elegant solution :D

Issue #994, opened 2017-04-10
Make String conform to HeaderFooterViewRepresentable


Sweeper777:
I want to do this:

        form +++ TextAreaRow("some tag") {
            row in
            row.section?.header = "Hello"
        }

But there is a compiler error around "some tag", saying that there is an extra argument. I think this is because Swift can't understand that it can convert the string literal to a `HeaderFooterView`, which is compatible with the required type `HeaderFooterViewRepresentable`. It would be better if `String` itself conformed to `HeaderFooterViewRepresentable`

m-revetria:
Hi @Sweeper777, thanks for the report.

Actually you cannot create a `HeaderFooterViewRepresentable` directly with a string literal. There is a mistake in the documentation where it says so.

As the generic class `HeaderFooterView` conforms to `ExpressibleByStringLiteral` you are able to create an instance of it from a string really simple. You can do something like below:

```swift
section.header = HeaderFooterView<UIView>(stringLiteral: "Header title here")
// or
let header: HeaderFooterView<UIView> = "Header title here"
section.header = header
```

We are not currently planning to make `String` conform to `HeaderFooterViewRepresentable` in the Eureka's project. You can make it in a simple way like it's shown below:

```swift
extension String: HeaderFooterViewRepresentable {

    public var title: String? {
        get {
            return self
        }
        set(newValue) { }
    }

    public var height: (() -> CGFloat)? {
        get { return nil }
        set { }
    }

    public func viewForSection(_ section: Section, type: HeaderFooterType) -> UIView? {
        return nil
    }

}
```

Another approach would be adding a computed property to Section through an extension:


```swift
extension Section {

    var headerString: String? {
        get { return header?.title }
        set {
            let header = HeaderFooterView<UIView>.init(title: newValue)
            self.header = header
        }
    }

}
```

Regards

Sweeper777:
Ok, thanks for the info!

PR #1010, opened 2017-04-20
Update README (remove section header string literal

allaire:
It's not working anymore (probably a leftover in the doc)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1021,8 +1021,6 @@ Look at this [issue](https://github.com/xmartlabs/Eureka/issues/96).
 * Set up a new header/footer data ....
 
 ```swift
-section.header = "Header Title" // use string literal as a header/footer data. HeaderFooterView conforms to ExpressibleByStringLiteral.
-//or
 section.header = HeaderFooterView(title: "Header title \(variable)") // use String interpolation
 //or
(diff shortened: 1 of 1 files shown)


m-revetria:
Related to #994

README.md
```diff
@@ -1021,8 +1021,6 @@ Look at this [issue](https://github.com/xmartlabs/Eureka/issues/96).
 * Set up a new header/footer data ....
 
 ```swift
-section.header = "Header Title" // use string literal as a header/footer data. HeaderFooterView conforms to ExpressibleByStringLiteral.
```
m-revetria: Actually you can create a `HeaderFooterView` instance from a string literal. So, maybe the deleted code can be replaced with:

```swift
let header: HeaderFooterView<UIView> = "Header Title" // use string literal as a header/footer data. HeaderFooterView conforms to ExpressibleByStringLiteral.
section.header = header
```

What do you think?

allaire: I think the exemple below is a better approach, and already documented:

```
section.header = HeaderFooterView(title: "Header title")
```

allaire: @m-revetria I can still add it if you prefer :)

m-revetria: @allaire sounds good to me. Thanks

PR #1028, opened 2017-04-28
Fixed how sections and row are inserted using subscripts

### LINKED ISSUE #1024: Section index, Row indexPath not set properly unless sections are "appended"
Environment: current Eureka, Xcode 8.3.2, iOS 10

FWIW, I have solved this problem, but I stumbled into what is either a bug or a very important documentation caveat. BaseCell.reload() and Section.reload() do not work unless Sections are  appended using +++ or form.append().

I have been developing my forms in sections, with sections added and replaced using `form[index] = section`. This **appears** to work properly, but does not call Form.append() under the hood, which means Section.wasAddedTo() and Row.wasAddedTo() are never called, so the internal values for index and indexPath are never set. 

I can't say for sure when I started using form[0] and section[0]-- I think I encountered it in sample code posted here. But it is clearly wrong, at least without some additional operator overloading.

m-revetria:
How it's pointed out in this issue #1024, if either sections or rows are added to the form using subscripts then they wont be correctly handled. For example `form[0][0].reload()` won't work if the rows was added in this way `form[0][0] = SomeRow()`

The changes fix how the sections and rows are inserted into the form, key points are:
* add sections and rows to `_allSections` and `_allRows` respectively
* call `wasAddedTo()` to each section and row added
* call `willBeRemovedFromForm` to each of the section and row that are being replaced.

Previous changes made subscripts works in the same way that `append` does

Fixes #1024

### DIFF
--- a/Source/Core/BaseRow.swift
+++ b/Source/Core/BaseRow.swift
@@ -212,6 +212,11 @@ extension BaseRow {
         removeFromRowObservers()
     }
 
+    final func willBeRemovedFromSection() {
+        willBeRemovedFromForm()
+        section = nil
+    }
+
     final func removeFromHiddenRowObservers() {
         guard let h = hidden else { return }
         switch h {
--- a/Source/Core/Form.swift
+++ b/Source/Core/Form.swift
@@ -172,7 +172,24 @@ extension Form: MutableCollection {
 
     public subscript (_ position: Int) -> Section {
         get { return kvoWrapper.sections[position] as! Section }
-        set { kvoWrapper.sections[position] = newValue }
+        set {
+            if position > kvoWrapper.sections.count {
+                assertionFailure("Form: Index out of bounds")
+            }
+
+            if position < kvoWrapper.sections.count {
+                let oldSection = kvoWrapper.sections[position]
+                let oldSectionIndex = kvoWrapper._allSections.index(of: oldSection as! Section)!
+                // Remove the previous section from the form
+                kvoWrapper._allSections[oldSectionIndex].willBeRemovedFromForm()
+                kvoWrapper._allSections[oldSectionIndex] = newValue
+            } else {
+                kvoWrapper._allSections.append(newValue)
+            }
+
+            kvoWrapper.sections[position] = newValue
+            newValue.wasAddedTo(form: self)
+        }
     }
     public func index(after i: Int) -> Int {
         return i+1 <= endIndex ? i+1 : endIndex
--- a/Source/Core/Section.swift
+++ b/Source/Core/Section.swift
@@ -193,6 +193,7 @@ open class Section {
 
     // MARK: Private
     lazy var kvoWrapper: KVOWrapper = { [unowned self] in return KVOWrapper(section: self) }()
+    
     var headerView: UIView?
     var footerView: UIView?
     var hiddenCache = false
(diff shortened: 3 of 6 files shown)


Source/Core/Section.swift
```diff
@@ -211,12 +211,29 @@ extension Section : MutableCollection, BidirectionalCollection {
             }
             return kvoWrapper.rows[position] as! BaseRow
         }
-        set { kvoWrapper.rows[position] = newValue }
+        set {
+            kvoWrapper.rows[position] = newValue
+            if position < kvoWrapper._allRows.count {
+                // Remove the previous row from the form
+                kvoWrapper._allRows[position].willBeRemovedFromForm()
```
mats-claassen: Is position the correct index for `_allRows`?

m-revetria: Yes I think so, why is the question? Rows are being indexed in the same way in both properties `rows` and `_allRows`, right?

mtnbarreto: _allRows contains hidden and non-hidden rows so index can be different at some point.

Source/Core/Form.swift
```diff
@@ -172,7 +172,17 @@ extension Form: MutableCollection {
 
     public subscript (_ position: Int) -> Section {
         get { return kvoWrapper.sections[position] as! Section }
-        set { kvoWrapper.sections[position] = newValue }
+        set {
+            kvoWrapper.sections[position] = newValue
+            if position < kvoWrapper._allSections.count {
+                // Remove an existing section from the form
+                kvoWrapper._allSections[position].willBeRemovedFromForm()
+                kvoWrapper._allSections[position] = newValue
```
mtnbarreto: S1, S2, S5                            <-- S6 index is 3

S1, S2, S3(H), S4(H), S5    <-- S6 index is 5 here

Source/Core/Form.swift
```diff
@@ -172,7 +172,23 @@ extension Form: MutableCollection {
 
     public subscript (_ position: Int) -> Section {
         get { return kvoWrapper.sections[position] as! Section }
-        set { kvoWrapper.sections[position] = newValue }
+        set {
+            if position > kvoWrapper.sections.count {
+                assertionFailure("Form: Index out of bounds")
+            }
+
+            if position < kvoWrapper._allSections.count {
```
mats-claassen: Shouldn't it be `sections` here instead of `allSections`?

mats-claassen: I think the code t=should be the same as in Section

mats-claassen: I still think this should be `sections` here

mtnbarreto: `sections` must be used there.

mtnbarreto:
@m-revetria please amend these commits and rebase if needed.

m-revetria:
@mtnbarreto I added several tests related to the changes introduced in this PR. Most of them wasn't passing before the changes, as we expected. All of them start passing with the new commits.

As an additional note, take a look to the change done in the section: when a row is being removed from a section a new function is called `willBeRemovedFromSection` on the row which will null its section reference.

Source/Core/BaseRow.swift
```diff
@@ -212,6 +212,11 @@ extension BaseRow {
         removeFromRowObservers()
     }
 
+    final func willBeRemovedFromSection() {
```
m-revetria: @mtnbarreto my thought when i did this change: if you are removing a row from a section then it will be removed from both the form and the section. But if you are removing a section from a form, then its rows will only be removed from the form but they will stay in the section. Let me know what do you think

Source/Core/Form.swift
```diff
@@ -172,7 +172,24 @@ extension Form: MutableCollection {
...
         get { return kvoWrapper.sections[position] as! Section }
-        set { kvoWrapper.sections[position] = newValue }
+        set {
+            if position > kvoWrapper.sections.count {
+                assertionFailure("Form: Index out of bounds")
+            }
+
+            if position < kvoWrapper.sections.count {
+                let oldSection = kvoWrapper.sections[position]
+                let oldSectionIndex = kvoWrapper._allSections.index(of: oldSection as! Section)!
```
mtnbarreto: @m-revetria  Do we actually need to case oldSection to Section?

nemecence90:
I get Assertion Failure when removing a row from the section, and only when I override rowsHaveBeenRemoved function. Let me know what I am doing wrong here?

nemecence90:
I figured it out, and I was too lame. I needed to call super on rowsHaveBeenRemoved and it took me one hour to figure it out. Thanks.

PR #1051, opened 2017-05-19
Make segmented cell more reusable

letko-dmitry:
Subj. Avoid cast to `SegmentedRow` type and cast to more general type: `OptionsRow`. It's more general and allow us to reuse cell if it's needed.


letko-dmitry:
Nope. It won't work. Sorry.

PR #1053, opened 2017-05-20
Added option to use accessory view in ImageCheckCell

ilyapuchka:
This PR allows to put image in ImageCheckCell in the accessory view to the right side of the title.


ilyapuchka:

PR #1055, opened 2017-05-20
Implemented lazy options provider for SelectorViewController

ilyapuchka:
This PR adds option to implement custom data provider for options selector, i.e. when options need to be fetched from remote source. After options are fetched they are cached unless provider is not reset to `nil`, which cleans the cash.


ilyapuchka:

mtnbarreto:
Closing this in favor of #1157

PR #1056, opened 2017-05-20
Fixed single selection for sectioned selector view controller

ilyapuchka:
This pr fixes the issue when selecting option in single option selector does not clear selection of previous value from another section.


ilyapuchka:

m-revetria:
Hi @ilyapuchka thanks for your contribution! Can you elaborate more on the issue that this PR is fixing? Maybe you can share a GIF with the bug to better understand what the proposed changes are doing.

ilyapuchka:
@m-revetria the bug manifests itself when `dismissOnChange = false` and `dismissOnSelection = false` so the options screen stays instead of pop and option in a different section is selected. In this case selection in other section is not cleared.
simulator screen shot 29 jun 2017 16 27 05

PR #1057, opened 2017-05-20
Improved various cells accessibility

ilyapuchka:
This PR improves accessibility of various rows types, i.e. text filed row for which currently the label is not read by VoiceOver if the value is already set. That makes it hard to understand what should be entered in the field.


ilyapuchka:

PR #1059, opened 2017-05-21
Custom SelectorViewController headers and footers

ilyapuchka:
This PR adds an option to change default headers in options selector which is currently built with system style headers and to change this requires subclassing and changing headers manually after form is setup.


anlaital:
+1 from me, as I need this feature also!

ilyapuchka:

PR #1064, opened 2017-05-23
Fixed #942

### LINKED ISSUE #942: SliderRow not disabled
- Environment: Eureka (master), Xcode 8.2.1 and iOS 10.2.1.

When I try and disable a SliderRow, the control is not disabled.  I'm disabling the row like this:

```swift
            let disabled = Condition.function(["disabled"], { (form) -> Bool in
                return true
            })

            row.disabled = disabled
            row.evaluateDisabled()
```

This technique works with Name, Stepper and Switch rows.

fedeojeda95:
Fixes the issue #942. The slider was not being disabled when the `SliderRow` was set disabled. This was due to the missing check if the row is disabled in the cell update

### DIFF
--- a/Source/Rows/SliderRow.swift
+++ b/Source/Rows/SliderRow.swift
@@ -97,6 +97,7 @@ open class SliderCell: Cell<Float>, CellType {
         valueLabel.isHidden = !shouldShowTitle && !awakeFromNibCalled
         titleLabel.isHidden = valueLabel.isHidden
         slider.value = row.value ?? 0.0
+        slider.isEnabled = !row.isDisabled
     }
 
     func addConstraints() {


Source/Rows/SliderRow.swift
```diff
@@ -97,6 +97,11 @@ open class SliderCell: Cell<Float>, CellType {
         valueLabel.isHidden = !shouldShowTitle && !awakeFromNibCalled
         titleLabel.isHidden = valueLabel.isHidden
         slider.value = row.value ?? 0.0
+        if row.isDisabled {
+            slider.isEnabled = false
```
santiagofm: Maybe just `slider.isEnabled = !row.isDisabled` ?

Source/Rows/SliderRow.swift
```diff
@@ -116,6 +117,7 @@ open class SliderCell: Cell<Float>, CellType {
     }
 
     func valueChanged() {
+        guard !row.isDisabled else { return }
```
mtnbarreto: this method should not be called if slider is not enabled.

mtnbarreto: Do we need this check anyway? @fedeojeda95

fedeojeda95: Yep, thats right. If the slider is not enabled, then this method will never be called.

PR #1066, opened 2017-05-25
Minor documentation fixes and adding comments.

chefren:


chefren:
@NunoAlexandre 
I will remove comments that may be redundant (not redundante). Please explain which one is an apology? (appology is not a word) for code. The majority are not even comments but documentation for the functions.
The comment and the link posted have several spelling and grammar mistakes.
I also disagree in just refactoring function names, apart from obvious interface change and breaking of compatibility, functions can be doing complex things that can not be expressed in a few words. Documentation is always needed for users, specially with no access to the code. And if there is access to the code, adding a few high level comments, helps maintaining the code.

README.md
```diff
@@ -467,27 +467,24 @@ form +++
                 $0.placeholder = "Tag Name"
             }
         }
-        $0 <<< NameRow() {
```
chefren: This seems to come from the demo code, not explained in the readme, so removing to avoid confusion.

NunoAlexandre:
@chefren, my apologies for my grammar mistakes. I hope it didn't stop you from understanding the content. The comments you have removed in the latest commit look just right! About the second part of your comment: I nowhere said to just improve methods names. Methods of more than 6 lines surely require comments, otherwise the cognitive load is too much, specially if that is the norm across a project. But, which is what I was suggesting, if the code itself is refactored (i.e., big methods split into small ones, better names, etc), comments most likely become unecessary. It's better to cure the cause than the symptoms. Anyways, nice improvement, I hope it gets merged in! Cheers

README.md
mtnbarreto: Most of time we want to show a data entry row when a MultiRow section is added with insertion mode.

mtnbarreto: This is the default row, otherwise the user to enter the first name will have to tap add button to insert a NameRow.

README.md
```diff
@@ -578,6 +575,10 @@ If you want to validate the entire form (all the rows) you can manually invoke F
 
 Each row has the `validationErrors` property that can be used to retrieve all validation errors. This property just holds the validation error list of the latest row validation execution, which means it doesn't evaluate the validation rules of the row.
 
+#### Note on types
+
+As expected, the Rules must use the same types as the Row object. Be extra careful to check the row type used. You might see a compiler error ("Incorrect arugment label in call (have 'rule:' expected 'ruleSet:')" that is not pointing to the problem when mixing types.
+
```
mtnbarreto: 👍

PR #1068, opened 2017-05-28
Fixing misspelling

feargswalsh92:
I fixed a misspelling of  MultivaluedOnlyReorderController I found on line 1533 of ViewController.swift, this is my first ever pull request so apologies in advance if I didn't do this properly but I read the Contributing guidelines about fifty times so I think I'm good.


Example/Example/ViewController.swift
```diff
@@ -1530,7 +1530,7 @@ class MultivaluedController: FormViewController {
     }
 }
 
-class MultivaluedOnlyRearderController: FormViewController {
```
m-revetria: This change must be reflected in the Storyboard where this class is used.

Eureka.xcworkspace/contents.xcworkspacedata
```diff
@@ -10,4 +10,7 @@
    <FileRef
       location = "group:Example.xcodeproj">
    </FileRef>
+   <FileRef
```
m-revetria: Please revert these changes as they are not related with the proposed change.

Example/Example/ViewController.swift
mtnbarreto: Will be fixed in Eureka Xcode 9 compatible version.

mtnbarreto:
@feargswalsh92 Thanks for reporting the issue. We will fix it in Xcode 9 compatible version that will be released shortly.

PR #1069, opened 2017-05-29
AnyHashable section keys for SelectorViewController

ilyapuchka:
This makes sectioned selectors a bit more flexible, but it looks like it is a breaking change. We might keep the old version of these closures and rename new, let me know what is more preferable


ilyapuchka:

PR #1073, opened 2017-05-30
Improved logic to allow changing the default cell height easier.

anlaital:
This makes it possible to easily change the default cell height for all cells by changing the `tableView.rowHeight` property on `FormViewController`.

The reason why this did not previously work was that `height` was always set in the base `Cell` implementation.

Please note that this change shouldn't affect pickers; they are still using `UITableViewAutomaticDimension`. This goes for inline pickers as well.

I tested this with the Example app included in the workspace and it worked for all cells present.


anlaital:
Any chance of getting this reviewed?

PR #1074, opened 2017-05-31
Fix compiler warnings.

anlaital:
This fixes all compiler warnings from both Eureka and the Example project.


m-revetria:
Thanks for your contribution!

PR #1087, opened 2017-06-08
Swift 4 migration

mcousillas6:
Migrated the code to support swift-4. The code is currently building successfully in Xcode9-beta.

*Known issues*

* `SelectableSections`  are not working properly in multiple-select mode. The first time that the cell is selected it works as expected but after that, it's not able to change value.


Source/Core/Section.swift
```diff
@@ -232,8 +232,8 @@ extension Section : MutableCollection, BidirectionalCollection {
         }
     }
 
-    public subscript (range: Range<Int>) -> [BaseRow] {
-        get { return kvoWrapper.rows.objects(at: IndexSet(integersIn: range)) as! [BaseRow] }
+    public subscript (range: Range<Int>) -> ArraySlice<BaseRow> {
+        get { return ArraySlice<BaseRow>(kvoWrapper.rows.objects(at: IndexSet(integersIn: range)) as! [BaseRow]) }
```
mtnbarreto: @mcousillas6 @m-revetria  I think ArraySlice should keep index from original collection, in this case the range. This array slice start and end index are 0 . number of items in the collection instead of range.startIndex, range.EndIndex.

mtnbarreto: I think this is the right migration.   
```swift
get { return kvoWrapper.rows.map({ $0 as! BaseRow })[range.lowerBound...range.upperBound] }
```

Maybe there is a better implementation.

mtnbarreto:
I will close this pull requestin in favor of my Xcode9-swift3_2 and Xcode9-Swift4 branches.

PR #1098, opened 2017-06-12
Merge two forms together with (+++)  operator

HusamAamer:
Operator (+++) should be able to add more than one section (Whole Form) to the left form

### DIFF
--- a/Source/Core/Operators.swift
+++ b/Source/Core/Operators.swift
@@ -109,6 +109,21 @@ public func +++ (left: BaseRow, right: BaseRow) -> Form {
     return form
 }
 
+/**
+ Merges two forms together
+ 
+ - parameter left:  form 1
+ - parameter right: form 2
+ */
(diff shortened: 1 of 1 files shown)


mtnbarreto:
Hi @HusamAamer ! Why do we need these operator? A section should only belongs to one form by definition. 

If for some reason this helps you in anything you can easily add this operator in your own application code. 

Let me know your thoughts..

HusamAamer:
Yes, I've added this in my code.
But I thought it would be nice if this operator was built in.

My usage was like this ..
For delivery app sign up form
```
let commonSignUpFields = ....
let driver_form = driver_only_sections
let restaurant_form = restaurant_only_sections

form +++ commonSignUpFields +++ (userType == .Driver ? driver_form : restaurant_form) 
```

anyway , thanks a lot for this gorgeous Eureka

PR #1113, opened 2017-06-19
Allow table view delegate methods to be overridden in subclasses.

sgerardi:


m-revetria:
@sgerardi merged. Thanks for your contribution!

Issue #1076, opened 2017-06-01
MultivaluedSection did remove row callback?


pluckq:
Is there any way to get that?

Laptopmini:
+1

macrigiuseppe:
+1

fabiodursoexin:
+1

mtnbarreto:
Take a look at https://github.com/xmartlabs/Eureka/pull/1136

stonkeep:
See on https://stackoverflow.com/questions/48916511/why-cant-delete-row-in-eureka-demo

Issue #998, opened 2017-04-11
How do I take custom actions when a row is removed from a multivalued section?


jfjdud120:
Eureka: 3.0.0
Xcode: 8.3

Hi.

For example, I want to get a reference to a row and print the title of it when it is removed from a multivalued section by swiping and tapping "delete" button.

In earlier version, I overrode `tableView(_ tableView: UITableView, commit editingStyle: UITableViewCellEditingStyle, forRowAt indexPath: IndexPath)` and could do what I wanted.

But, in Eureka 3.0.0, I can't override it, because it is not 'open'.

If I can't override it, there must be some mechanism for doing what I want(e.g. delegate or callback, etc.), I guess.

I found a method `rowsHaveBeenRemoved(_ rows: [BaseRow], at indexes: [IndexPath])`, but Xcode says 'Declarations from extensions cannot be overridden yet.'

How can I do that?

mtnbarreto:
Let's make the method open for now. I think it doesn't worth adding a new callback just to get multivalued section notifications.

Laptopmini:
Bump.

macrigiuseppe:
was this resolved?

mtnbarreto:
For a solution take a look at #1136

stonkeep:
See on https://stackoverflow.com/questions/48916511/why-cant-delete-row-in-eureka-demo

PR #1136, opened 2017-07-07
Added delete callback for RowType instances

AdamStreet:
Named as `willBeRemoved` 

Referred issues:

* https://github.com/xmartlabs/Eureka/issues/998
* https://github.com/xmartlabs/Eureka/issues/1076

### DIFF
--- a/Source/Core/BaseRow.swift
+++ b/Source/Core/BaseRow.swift
@@ -27,6 +27,7 @@ import Foundation
 open class BaseRow: BaseRowType {
 
     var callbackOnChange: (() -> Void)?
+	var callbackWillBeRemoved: (() -> Void)?
     var callbackCellUpdate: (() -> Void)?
     var callbackCellSetup: Any?
     var callbackCellOnSelection: (() -> Void)?
--- a/Source/Core/Core.swift
+++ b/Source/Core/Core.swift
@@ -776,6 +776,7 @@ extension FormViewController : UITableViewDelegate {
             if let _ = row.baseCell.findFirstResponder() {
                 tableView.endEditing(true)
             }
+			row.callbackWillBeRemoved?()
             section.remove(at: indexPath.row)
             DispatchQueue.main.async {
                 tableView.isEditing = !tableView.isEditing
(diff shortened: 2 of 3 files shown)


SkylerSeamans:
This would be incredibly helpful! Hope it gets a review soon.

Raymond-em:
This is will be very useful, also a similar callback to reordering of rows

mtnbarreto:
Hi @AdamStreet! 
I'm not sure that we need this callback to get notifications about multivalued rows deletion. 

I would prefer a more semantic name and also expose the property from another class, probably MultivaluedSection so we make it clear that the closure only works for multivalued sections. With the current implementation, it seems that the `willBeRemoved ` is called each time a row is removed from the form which is wrong. 

An alternative solution could be:

The `Section` class conforms to `SectionDelegate` so implements the following methods:

/// The delegate of the Eureka sections.
public protocol SectionDelegate: class {
    func rowsHaveBeenAdded(_ rows: [BaseRow], at: IndexSet)
    func rowsHaveBeenRemoved(_ rows: [BaseRow], at: IndexSet)
    func rowsHaveBeenReplaced(oldRows: [BaseRow], newRows: [BaseRow], at: IndexSet)
}

So you can extend MultivaluedSection to override these methods (do not forget to call super implementation). 

> Anyway I'm gonna move the `FormDelegate` methods from extension to `FromViewController` class to make them extensible.

AdamStreet:
Hi @mtnbarreto,

Thank you for reviewing my request. 
The reason why I chose to add this callback - even the naming is not optimal and also works on Rows of `Section` instances - instead of overriding the delegate method(s) was the `-onChange` method syntax. If you handle the _value change_ event in an expression it was just as convenient to do so when the row was or will be removed. 

Can you please describe a scenario when the current implementation causes "wrong" behaviour? 

> With the current implementation, it seems that the `willBeRemoved` is called each time a row is removed from the form which is wrong.

If the user removes a row by hitting the *delete* button it fires the callback - I haven't tested if it was done programatically but I would expect the same when that happens.

I haven't found any issue using this implementation but I am happy to switch if it is unreliable or has a hidden flaw inside.

yerzhanserikbay:
public protocol SectionDelegate: class {
    func rowsHaveBeenAdded(_ rows: [BaseRow], at: IndexSet)
    func rowsHaveBeenRemoved(_ rows: [BaseRow], at: IndexSet)
    func rowsHaveBeenReplaced(oldRows: [BaseRow], newRows: [BaseRow], at: IndexSet)
}

Please, can you show me an example ?

PR #1148, opened 2017-07-13
change validation on demand behaviour when row was not valid

bennibau:
Validating on demand should ONLY validate when u demand it and not, like it is now, always when something changes. I need this in a concrete project because I get strange validation behaviour if it validates all the time.

### DIFF
--- a/Source/Core/Row.swift
+++ b/Source/Core/Row.swift
@@ -55,7 +55,7 @@ open class RowOf<T: Equatable>: BaseRow {
             _value = newValue
             guard let _ = section?.form else { return }
             wasChanged = true
-            if validationOptions.contains(.validatesOnChange) || (wasBlurred && validationOptions.contains(.validatesOnChangeAfterBlurred)) || !isValid {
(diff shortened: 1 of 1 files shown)


Source/Core/Row.swift
```diff
@@ -55,7 +55,7 @@ open class RowOf<T: Equatable>: BaseRow {
             _value = newValue
             guard let _ = section?.form else { return }
             wasChanged = true
-            if validationOptions.contains(.validatesOnChange) || (wasBlurred && validationOptions.contains(.validatesOnChangeAfterBlurred)) || !isValid {
+            if validationOptions.contains(.validatesOnChange) || (wasBlurred && validationOptions.contains(.validatesOnChangeAfterBlurred)) || (!isValid && !validationOptions.contains(.validatesOnDemand)) {
```
mtnbarreto: For me the right code should be:

`if validationOptions.contains(.validatesOnChange) || (wasBlurred && validationOptions.contains(.validatesOnChangeAfterBlurred)) || (!isValid && validationOptions != .validatesOnDemand) `

because if you have `validatesOnBlur` or any other validation option and the row is not valid you want to validate it whenever a  value change happens.

bennibau:
I accepted and implemented the changes

PR #1195, opened 2017-08-17
Make Inline Cell Visibility Animation After Expansion Conditional

enhorn:
This adds a toggle to enable avoiding centering on expanded inline cells automatically.
The default value is `true` so the behavior stays the same as it is now as default.

Why I need it in my project:
I have an embedded `FormViewController` inside a `UITableViewCell` that auto sizes it self for changes in the content size on contained scrollviews. The auto positioning is causing layout issues for me in this scenario.


MrAlek:
Could someone look at merging this? @mtnbarreto @mats-claassen 🙏

Issue #1178, opened 2017-08-01
can we add #if swift(>=3.2) to swift branch to pull into master


jaz-ah:
Ideally instead of having to point my Pods to your swift feature branches, we would add #if swift conditionals in the code to compile correctly on either Xcode8.3.3 or Xcode9.0 - then you could release a new version that compiles for either and we wouldn't have to have feature branches to point to different pod versions in our own codebase.... @mtnbarreto thoughts?

jaz-ah:
for example in Section.swift:

    #if swift(>=3.2)
    public subscript (range: Range<Int>) -> ArraySlice<BaseRow> {
        get { return kvoWrapper.rows.map({ $0 as! BaseRow })[range.lowerBound...range.upperBound] }
        set {
            replaceSubrange(range, with: newValue)
        }
    }
    #else
    public subscript (range: Range<Int>) -> [BaseRow] {
        get { return kvoWrapper.rows.objects(at: IndexSet(integersIn: range)) as! [BaseRow] }
        set {
            replaceSubrange(range, with: newValue)
        }
    }
    #endif

mtnbarreto:
Hi @jaz-ah , Once Xcode 9 gets released, the latest Eureka release source code will be in swift 4. According Swift documentation we can continue having our code in swift 3 and some frameworks (including Eureka) in swift 4 due to swift 4 compatibility capabilities.

jaz-ah:
@mtnbarreto understood, but I'm wondering until then if we can make master at least swift3.2 compatible w/ the #if's above?

ldiqual:
@mtnbarreto My understanding is that the Swift 3.2 branch does not compile on Xcode 8, which means our devs have to use the Swift 3.2 compiler as soon as our app bumps the eureka dependency. Is conditional compiling for <3.2 something that is into consideration? Any alternative maybe?

jaz-ah:
y - simplest changes are to conditionally compile like my example above which is compatible w/ both xcode8+9 and means devs don't have to have different branches/etc... while we're waiting for xcode9/swift4 to be final yet...

mspasov:
@jaz-ah is right, it's very convenient to have Xcode 8 & Xcode 9 compatibility in the pre-swift 4 branch . Swift compiler conditionals are a safe bet. This would allow working on the project with Xcode9, but compiling it for App Store release with Xcode 8.3. I can prepare a PR for this.

ldiqual:
[...]

PR #1199, opened 2017-08-24
Add support for Swift 3.1 in the Swift 3.2 branch

ldiqual:
Fixes https://github.com/xmartlabs/Eureka/issues/1178

As described in the issue above, it would be great to support both the Swift 3.1 and 3.2 compilers, at least for a little while since people will slowly transition from Xcode 8 to Xcode 9. When merged, this will allow people to develop for iOS 11 on the Xcode 9 betas all while publishing with Xcode 8 or AppCode.

Summary of changes:

 - Added back the old signature for `replaceSubrange` in `Section` and put it behind a version check. I'm not sure if there's a way to consolidate the code between the swift 3.1/3.2 methods so I'm open to suggestions.
 - Added a `_ImplicitlyHashable` type which works around a warning in the 3.2 compiler for redundant generic constraints. The workaround is described [in the code](https://github.com/xmartlabs/Eureka/pull/1199/files#diff-83e01711b59e6f7f5bad32b208fcaafeR82).

I tested the code + tests + app both in Xcode 8 and Xcode 9 beta 5. Everything compiles/runs without warnings.

Let me know if you have any questions!

### DIFF
--- a/Source/Core/Helpers.swift
+++ b/Source/Core/Helpers.swift
@@ -79,6 +79,12 @@ extension NSExpression {
     }
 }
 
+// This is a workaround for warnings in the Swift 3.2 compiler that are triggered when a generic type
+// is explicitly constrained while already implicitly constrained.
+// For instance `T: Hashable where Cell.Value == Set<T>` triggers a warning on Swift 3.2 because Set already
+// has the constraint `T: Hashable`.
+// However Swift 3.1 doesn't infer this constraint, so we need to constraint explicitly. To do this,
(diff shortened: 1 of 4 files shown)


jaz-ah:
+1 LGTM - thx @ldiqual

ldiqual:
@mtnbarreto Would you mind taking a look at this PR? Thank you

mtnbarreto:
I would prefer to write all Eureka code in swift 4. Xcode 9 swift 3.2 can still use it without any problem. 

swift 3.1 users will have to use a older Eureka version that supports it. 

I'm still gonna merge these pr but it won't be merged to master.

PR #1214, opened 2017-09-08
Fix nil headers/footers of sections are displayed in FormViewController in Xcode 9

banjun:
https://download.developer.apple.com/Documentation/Beta_Release_Notes_Sep_6_2017/iOS_11_beta_10_Release_Notes.pdf

As noted in iOS 11 release note, the default values of `estimatedSectionHeaderHeight` and `estimatedSectionFooterHeight` have changed in Xcode 9. This cause empty section headers/footers to be displayed with non-zero height even for header/footer-less sections.
To preserve iOS 10 SDK behavior, I inserted code that overrides the default values right after FormViewController creates UITableView.

This might not occur when FormViewController is initialized by a storyboard because default values are ignored by storyboards. Initializing by code should reproduce this issue.

A minimum reproducible example usecase is here:

https://github.com/banjun/EurekaXcode9/blob/e55a6465ef49f42559702e82dcef038011f7dcac/EurekaXcode9/ViewController.swift

### DIFF
--- a/Source/Core/Core.swift
+++ b/Source/Core/Core.swift
@@ -463,6 +463,10 @@ open class FormViewController: UIViewController, FormViewControllerProtocol {
             if #available(iOS 9.0, *) {
                 tableView.cellLayoutMarginsFollowReadableWidth = false
             }
+            // in Xcode 9, with plain style, estimated section header/footer heights are automatic by default and measure non-zero even for header/footer-less sections.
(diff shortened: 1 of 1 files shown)


banjun:
We can attach screenshots as Xcode 9 is now public.

before patch (nil section headers/footers display in plain style):


after patch:

mtnbarreto:
I understand the problem and it may break other use cases. If for some reason to experience this wrong height header calculation you can set up it from your FormViewController extension class.

PR #1215, opened 2017-09-08
MultivaluedSection.values()

bmichotte:
This PR adds the values of a `MultivaluedSection` into `form.values()`. This will fix #1015 and fix #981 

Tests have been updated, the swift 4 port is also available in next PR

capture d ecran 2017-09-08 a 09 37 33

PR #1216, opened 2017-09-08
MultivaluedSection.values() swift4

bmichotte:
This is the swift 4 port of #1215

### DIFF
--- a/Source/Core/Form.swift
+++ b/Source/Core/Form.swift
@@ -106,19 +106,11 @@ public final class Form {
      */
     public func values(includeHidden: Bool = false) -> [String: Any?] {
         if includeHidden {
-            return allRows.filter({ $0.tag != nil })
-                .reduce([String: Any?]()) {
-                    var result = $0
-                    result[$1.tag!] = $1.baseValue
-                    return result
-                }
+            return getValues(for: allRows.filter({ $0.tag != nil }))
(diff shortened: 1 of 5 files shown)


Source/Core/Section.swift
```diff
@@ -470,4 +470,13 @@ open class MultivaluedSection: Section {
         }
         self <<< addRow
     }
+
+    /**
+     Method used to get all the values of the section.
+
+     - returns: An Array mapping the row values. [value]
+     */
+    public func values() -> [Any?] {
```
mtnbarreto: why does it public?

Source/Core/Form.swift
```diff
@@ -360,6 +352,22 @@ extension Form {
         }
         kvoWrapper.sections.insert(section, at: formIndex == NSNotFound ? 0 : formIndex + 1 )
     }
+
+    func getValues(for rows: [BaseRow]) -> [String: Any?] {
```
mtnbarreto: could it be private?

Source/Core/Section.swift
mtnbarreto: I would prefer to remove this method

bmichotte: This way, you can call 
```
let section = form.sectionBy(tag: "sectionTag")
section.values()
```

Source/Core/Form.swift
bmichotte: indeed

PR #1229, opened 2017-09-19
fixed deallocating section

### LINKED ISSUE #1218: Selectable list sections aren't selectable in Xcode 9/Swift 4
Using the latest build of Xcode 9/iOS 11 and Eureka's Xcode 9/Swift 4 branch.

Selectable list sections don't seem to work in the most recent Xcode 9 build. When you select the first option it works fine, but after selecting further ones they only update occasionally to being checked. (From what I can tell, when choosing a value the first time it'll work, but you can't choose the same value twice if you say, change your mind.)

If I create a simple "Single View App" project, and have this as `ViewController.swift` (code taken from the example in the README):

````swift
import UIKit
import Eureka

class ViewController: FormViewController {
    override func viewDidLoad() {
        super.viewDidLoad()
        
        view.backgroundColor = .green
        
        form +++ SelectableSection<ListCheckRow<String>>("Where do you live", selectionType: .singleSelection(enableDeselection: true))
        
        let continents = ["Africa", "Antarctica", "Asia", "Australia", "Europe", "North America", "South America"]
        for option in continents {
            form.last! <<< ListCheckRow<String>(option){ listRow in
                listRow.title = option
                listRow.selectableValue = option
                listRow.value = nil
            }
        }
    }
}
````

The behaviour can be seen in that code in Xcode 9. In Xcode 8 it works properly.

### PR DESCRIPTION
**Pull request by ilyapuchka**
Resolves #1218 

### Explanation

In Swift 4 `filter` method called in this case has a [new](https://github.com/apple/swift-evolution/blob/master/proposals/0174-filter-range-replaceable.md) behaviour. Implementation of `RangeReplaceableCollection.filter` creates new array of the same type (`SelectableSection`) by calling `init` (can be checked by adding breakpoint at `SelectableSection.init` method) and then calling `append(contentsOf:)` method, which results in adding previously selected row (only this row will have non-nil value) to this section and so overriding a reference to it from `self` (in this scope).

So basically what happens is that every time raw is selected it creates new section with just this row. Then when we leave the stack this section is being deallocated as it is not referenced anywhere else and it's no more possible to select this row, because `guard` does not pass any more. Though current section (`self` in this context) still contains this row.

### DIFF
--- a/Example/Example/ViewController.swift
+++ b/Example/Example/ViewController.swift
@@ -239,7 +239,10 @@ class RowsExampleViewController: FormViewController {
                         $0.options = [💁🏻, 🍐, 👦🏼, 🐗, 🐼, 🐻]
                         $0.value = 👦🏼
                         $0.selectorTitle = "Choose an Emoji!"
-                    }
+                    }.onPresent { from, to in
+                        to.dismissOnSelection = false
+                        to.dismissOnChange = false
+            }
 
                 <<< PushRow<Emoji>() {
                         $0.title = "SectionedPushRow"
@@ -247,6 +250,8 @@ class RowsExampleViewController: FormViewController {
                         $0.value = 👦🏼
                         $0.selectorTitle = "Choose an Emoji!"
                     }.onPresent { from, to in
+                        to.dismissOnSelection = false
+                        to.dismissOnChange = false
                         to.sectionKeyForValue = { option in
                             switch option {
                             case 💁🏻, 👦🏼: return "People"
--- a/Source/Core/SelectableSection.swift
+++ b/Source/Core/SelectableSection.swift
@@ -89,7 +89,8 @@ extension SelectableSectionType where Self: Section {
                     case .multipleSelection:
                         row.value = row.value == nil ? row.selectableValue : nil
                     case let .singleSelection(enableDeselection):
-                        s.filter { $0.baseValue != nil && $0 != row }.forEach {
+                        s.forEach {
+                            guard $0.baseValue != nil && $0 != row else { return }
                             $0.baseValue = nil
                             $0.updateCell()
                         }

### CODE REVIEW

**Comment by ilyapuchka** [comment]
@mtnbarreto after #1230 and #1231 this bug is still happening, please can you take a look at this fix?

**Comment by mtnbarreto** [comment]
I may have others wrong filter method uses. 

This code also works. 
```swift
    /**
     Internal function used to set up a collection of rows before they are added to the section
     */
    func prepare(selectableRows rows: [BaseRow]) {
        for row in rows {
            if let row = row as? SelectableRow {
                row.onCellSelection { [weak self] cell, row in
                    guard let s = self, !row.isDisabled else { return }
                    switch s.selectionType {
                    case .multipleSelection:
                        row.value = row.value == nil ? row.selectableValue : nil
                    case let .singleSelection(enableDeselection):
                        let filterArray: [Self.Element] = s.filter { $0.baseValue != nil && $0 != row }
                        filterArray.forEach {
                            $0.baseValue = nil
                            $0.updateCell()
                        }
                        // Check if row is not already selected
                        if row.value == nil {
                            row.value = row.selectableValue
                        } else if enableDeselection {
                            row.value = nil
                        }
                    }
                    row.updateCell()
                    s.onSelectSelectableRow?(cell, row)
                }
            }
        }
    }
```

ilyapuchka:
Yes, because it will use different implementation of filter method. Even just adding type cast will make it work `(s.filter { $0.baseValue != nil && $0 != row } as [Self.Element])`, but I would prefer just to use `forEach` with `guard`. Your call though, I'm just glad I could figure out the issue eventually.

mtnbarreto:
@ilyapuchka There is an issue that whenever a row is added to a section, the row's `section` property  is updated with this new section (the section that swift 4 filter creates in this case).

I couldn't figured out why self becomes nil when filter method returns a new collection. 
`self` which is the selectable section should be referenced by the form .

This may occur in other parts of eureka.

mtnbarreto:
Got it. `func prepare(selectableRows rows: [BaseRow]) ` is called for the new filter created section.

ilyapuchka:
It becomes nil because when filter creates new collection (which is section in this case) and adds it's items that pass filter (rows) and then all the kvo-triggered side effects are performed and row gets new reference to section, as you pointed, but this section is not referenced anywhere, and reference in row is weak, so when we leave the context of this method it becomes nil, which makes total sense. All this is a new implementation of `filter` in `RangeReplaceableCollection` protocol that is only available for swift 4 (what explains why it works in swift 3). Changing type of return value of filter will change version of the function used here (it will get variant from `Sequence` protocol), which does not create new collection, but instead creates array, so it does not call `append` and reference to the section in the row is not overridden. 
The similar thing will happen if we filter form.

I went through all usages of filter and I think they are all fine, they are either performed on arrays, for which I believe behaviour didn't change, or their return type is explicitly set to array. There are some usages in `CollectionTests` but they don't seem to affect tests.

It would be better to avoid this behaviour as it can be confusing and I can easily imaging me as a consumer of the library to apply filter on the section, but I'm not sure how this method can be reimplemented avoiding side effects and still preserving return type, unless we also start to create new rows, not just section.

mtnbarreto:
Ideally either `filter` method that returns the Filtered associated type should not be used or the method should change the section itself by removing not included (non filtered) rows.

> We have the same issue for the Form.

Conceptually speaking, a row can be added to only one section and a section to only one form. Doesn't make sense to have the same row added to 2 sections since a row holds its view UITableViewCell. 

maybe, in order to avoid any usage confusion, we can redefine `filter` method and add some annotation that indicates to do not use it because its side effects.  What do you think?

ilyapuchka:
I like the first option, it should be just clearly documented that it will return the same section with removed rows. But we can also redefine this method and add deprecated annotation and fatal error instead of implementation or some basic implementation.

PR #1246, opened 2017-09-25
Allow non-selectable rows to exist besides selectable rows

dword-design:
From time to time it is helpful to have non-selectable rows like switch rows besides selectable rows (e.g. as inline rows). While this basically works, it fails when the selected row changes. In this case, also the values of the non-selectable rows are reset. With a little check this can be fixed.


mtnbarreto:
@dword-design thanks for contributing!!

PR #1106, opened 2017-06-15
iOS11 SwipeActions

marbetschar:
Containing a graceful fallback to the old behaviour using `editActionForRowAt: indexPath`. This covers https://github.com/xmartlabs/Eureka/issues/105.

PR #1249, opened 2017-09-26
Swift 4 swipeable

marbetschar:
According to https://github.com/xmartlabs/Eureka/pull/1106 this pull request now goes to the new swift 4 branch.


AAverin:
Right when I was wondering how to add swipeable actions – here is a PR =)
Would be great to have it merged

marbetschar:
@mtnbarreto can we get this merged?

marbetschar:
any news on this?

kronik:
waiting for this PR to be merged too

marbetschar:
@mtnbarreto whats missing to get this merged? I'd love merging this the swipe feature as soon as possible into the official release, since I need it in my current project. That said, it would be awesome to have this pull request either merged - or getting to know whats wrong or missing, so I can fix/add the stuff needed and to get this merged based upon the suggestions :)

marbetschar:
@mats-claassen or @mtnbarreto whats missing to get this swipe feature merged or declined?

Source/Core/SwipeActions.swift
```diff
@@ -0,0 +1,134 @@
...
+public class SwipeConfiguration{
+	
+	public init(configure: (SwipeConfiguration) -> Void){
+		configure(self)
+	}
+	
+	public var performsFirstActionWithFullSwipe: Bool = false
+	public var actions: [SwipeAction] = []
+	
+	public var platformValue: Any?{
```
m-revetria: Seems that this can be available just from iOS 11, right? If so then you can set its type to `UISwipeActionsConfiguration`

Source/Core/SwipeActions.swift
```diff
@@ -0,0 +1,134 @@
...
+				handler(action,indexPath,nil)
+			}
+		}
+	}
+	
+	public enum Style: Int{
+		case normal = 0
+		case destructive = 1
+		
+		var platformValue: Any{
```
m-revetria: Would be great if all these `platformValues` can have a type more concrete than `Any`. @mtnbarreto thoughts?

Example/Example/ViewController.swift
```diff
@@ -106,15 +106,11 @@ class HomeViewController : FormViewController {
...
-                
-        +++ Section()
-                <<< ButtonRow() { (row: ButtonRow) -> Void in
-                   row.title = "About"
-                }
-                .onCellSelection { [weak self] (cell, row) in
-                    self?.showAlert()
-                }
+			
+				<<< ButtonRow("Swipe Actions") { (row: ButtonRow) -> Void in
```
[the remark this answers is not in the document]
marbetschar: Done.

Example/Example/ViewController.swift
```diff
@@ -1747,6 +1743,64 @@ class MultivaluedOnlyDeleteController: FormViewController {
     }
 }
 
+class SwipeActionsController: FormViewController {
+	
+	override func viewDidLoad() {
+		super.viewDidLoad()
+		
+		form +++ Section(footer: "Eureka sets table.isEditing = false for SwipeActions.\n\nMultivaluedSections need table.isEditing = true, therefore both can't be used on the same view.")
+			<<< LabelRow("Actions Right: iOS >= 7"){
```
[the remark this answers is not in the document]
marbetschar: Done.

Source/Core/BaseRow.swift
```diff
@@ -98,7 +98,18 @@ open class BaseRow: BaseRowType {
     public var isHidden: Bool { return hiddenCache }
 
     /// The section to which this row belongs.
-    public weak var section: Section?
+    open weak var section: Section?
```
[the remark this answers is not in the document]
marbetschar: Right. Done.

Source/Core/SwipeActions.swift
```diff
@@ -0,0 +1,134 @@
...
+//
+
+import Foundation
+
+public typealias SwipeActionHandler = (Any, Any, ((Bool) -> Void)?) -> Void
+
+
+public class SwipeAction{
+	
+	public let platformValue: Any
```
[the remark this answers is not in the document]
marbetschar: Used `contextualAction` as suggested.

Source/Core/SwipeActions.swift
```diff
@@ -0,0 +1,134 @@
...
+	
+	public var backgroundColor: UIColor?{
```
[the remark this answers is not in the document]
marbetschar: Good point! Is changed.

Source/Core/SwipeActions.swift
```diff
@@ -0,0 +1,134 @@
...
+	public let style: Style
+	
+	public init(style: Style, title: String?, handler: @escaping SwipeActionHandler){
+		self.style = style
+		self.handler = handler
+		self.title = title
+		
+		if #available(iOS 11, *){
+			self.platformValue = UIContextualAction(style: style.platformValue as! UIContextualAction.Style, title: title){ action, view, completion -> Void in
+				handler(action,view,completion)
```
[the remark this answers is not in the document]
marbetschar: Done.

Source/Core/SwipeActions.swift
marbetschar: Good point. Used marker protocols to fix this. See `SwipeActions.swift` in the bottom of the file.

Source/Core/SwipeActions.swift
marbetschar: Good point! Fixed.

Source/Core/SwipeActions.swift
```diff
@@ -0,0 +1,168 @@
...
+//  Swipe.swift
+//  Eureka
+//
+//  Created by Marco Betschart on 14.06.17.
+//  Copyright © 2017 Xmartlabs. All rights reserved.
+//
+
+import Foundation
+
+public typealias SwipeActionHandler = (ContextualAction, ContextualActionSource, ((Bool) -> Void)?) -> Void
```
mtnbarreto: I think we should pass the row as closure argument.

Source/Core/SwipeActions.swift
```diff
@@ -0,0 +1,168 @@
...
+
+public class SwipeAction{
+	public let contextualAction: ContextualAction
+	
+	@available(iOS 11, *)
+	public var backgroundColor: UIColor?{
```
mtnbarreto: I would prefer to make `UIContextualAction` and `UITableViewRowAction ` conform to  a protocol like this:

```swift
public protocol ContextualAction{
     var backgroundColor: UIColor?
     var image: UIImage?
     var title: String?
}
```

Source/Core/SwipeActions.swift
```diff
@@ -0,0 +1,168 @@
...
+}
+
+
+public class SwipeConfiguration{
+	
+	public init(configure: (SwipeConfiguration) -> Void){
+		configure(self)
+	}
+	
+	public var performsFirstActionWithFullSwipe: Bool = false
```
mtnbarreto: remove `: Bool`

Source/Core/SwipeActions.swift
```diff
@@ -0,0 +1,168 @@
...
+				fatalError("Invalid platformValue for SwipeActionStyle: \(contextualStyle)")
+			}
+		}
+	}
+}
+
+
+public class SwipeConfiguration{
+	
+	public init(configure: (SwipeConfiguration) -> Void){
```
mtnbarreto: I would prefer a struct here and each row should have an instance of the strict already created so Eureka users can do something like that:

```
row.leftSwipe.actions = [swipeAction1, swipeAction2]
row.rightSwipe.actions = [swipeAction1, swipeAction2]
```

init should be refactored to `init()` in this case.

Source/Core/SwipeActions.swift
```diff
@@ -0,0 +1,168 @@
...
+			guard let contextualAction = self.contextualAction as? UIContextualAction else{ return nil }
+			return contextualAction.image
+		}
+		set{
+			guard let contextualAction = self.contextualAction as? UIContextualAction else{ return }
+			return contextualAction.image = newValue
+		}
+	}
+	
+	public var title: String?{
```
mtnbarreto: remove it since we can do `contextualAction.title`, `contextualAction.image` and `contextualAction. backgroundColor` and so on can be accessed directly using the protocol containing these properties.

Source/Core/SwipeActions.swift
mtnbarreto: `ContextualActionSource` can be removed since the row is passed as argument now.

Source/Core/SwipeActions.swift
```diff
@@ -0,0 +1,168 @@
...
+				switch self{
+				case .normal:
+					return UITableViewRowActionStyle.normal
+				case .destructive:
+					return UITableViewRowActionStyle.destructive
+				}
+			}
+		}
+		
+		public init(contextualStyle: ContextualStyle){
```
mtnbarreto: We don't need this init.

Source/Core/SwipeActions.swift
```diff
@@ -0,0 +1,168 @@
...
+			}
+			
+		} else {
+			self.contextualAction = UITableViewRowAction(style: style.contextualStyle as! UITableViewRowActionStyle,title: title){ (action, indexPath) -> Void in
+				handler(action, indexPath, nil)
+			}
+		}
+	}
+	
+	public enum Style: Int{
```
mtnbarreto: unnecessary `: Int` here.

Source/Core/SwipeActions.swift
```diff
@@ -0,0 +1,168 @@
...
+@available(iOSApplicationExtension 11.0, *)
+extension UIContextualAction: ContextualAction{}
+
+public protocol ContextualStyle{}
+extension UITableViewRowActionStyle: ContextualStyle{}
+
+@available(iOSApplicationExtension 11.0, *)
+extension UIContextualAction.Style: ContextualStyle{}
+
+public protocol ContextualActionSource{}
```
mtnbarreto: no longer needed according my feedback.

mtnbarreto:
@marbetschar We also need to add feature documentation in the readme in order to get this merged.

Source/Core/SwipeActions.swift
marbetschar: That's a brilliant idea! We should ideally use the specific Row type right? Not just `BaseRow`.

Source/Core/SwipeActions.swift
marbetschar: Of course 👍🏼 makes a whole lot more sense.

Source/Core/SwipeActions.swift
marbetschar: This looks amazing. One question: Would you prefer using `leftSwipe` and `rightSwipe` instead of `leadingSwipe` and `trailingSwipe` ?

If I'll look at `PushRow`, `left/right` looks more familiar. But due to the nature of this gesture it might lead to confusions.

marbetschar:
@mtnbarreto where to put the feature documentation?

Source/Core/SwipeActions.swift
marbetschar: Compiler throws `Enum case cannot have a raw value if the enum does not have a raw type` if `: Int` is removed

Source/Core/SwipeActions.swift
marbetschar: Implemented as `public typealias SwipeActionHandler = (ContextualAction, BaseRow, ((Bool) -> Void)?) -> Void`

Source/Core/SwipeActions.swift
marbetschar: Had to add the following in order to make `UITableViewRowAction` conform to the newly created `ContextualAction` protocol:

```
extension UITableViewRowAction: ContextualAction {
	public var image: UIImage? {
		get { return nil }
		set { return }
	}
}
```

Source/Core/SwipeActions.swift
marbetschar: Despite its true these properties can be accessed directly it looks a bit awkward in the form definition code. That's why I also made `SwipeAction` conforming to the newly created `ContextualAction` protocol - and let `SwipeAction.contextualAction` create the platform specific action on demand.

This way we end up with a nice and clean looking form initialisation code:

```
let infoAction = SwipeAction(style: .normal, title: "Info", handler: { (action, row, completionHandler) in
	print("Info")
	completionHandler?(true)
})
infoAction.backgroundColor = .blue

$0.leadingSwipe.actions = [infoAction]
$0.leadingSwipe.performsFirstActionWithFullSwipe = true
```

Source/Core/SwipeActions.swift
marbetschar: Removed.

Source/Core/SwipeActions.swift
marbetschar: Done using `leadingSwipe` and `trailingSwipe`.

Source/Core/SwipeActions.swift
marbetschar: Removed.

Source/Core/SwipeActions.swift
marbetschar: Removed `ContextualActionSource`

marbetschar:
@mtnbarreto also added feature documentation to README.md

mtnbarreto:
@marbetschar Finally it got merged! https://github.com/xmartlabs/Eureka/pull/1361. Thanks for contributing!

marbetschar:
@mtnbarreto perfect timing, thank you very much!!!! 🎉🎄🌟🎅🏻

matiasvillaverde:
Is possible to update this version by Cocoapods? Now the last version is 4.0.1

Thanks for this cool feature!

PR #1256, opened 2017-09-29
Disable iOS 11 section header/footer self-sizing

jeffaburt:
This fixes an iOS 11-specific issue where the table view would add space for section headers and footers, even if `tableView(_:viewForHeaderInSection:)` or `tableView(_:viewForFooterInSection:)` returned `nil`. This commit turns off self-sizing for section headers and footers, just like how we are already handling `tableView(_:estimatedHeightForRowAt:)`.

See [this Stack Overflow post](https://stackoverflow.com/a/46257601/1926015) for more info.

## Screenshot
slice

### DIFF
--- a/Source/Core/Core.swift
+++ b/Source/Core/Core.swift
@@ -787,6 +787,14 @@ extension FormViewController : UITableViewDelegate {
         }
         return view.bounds.height
     }
+    
+    open func tableView(_ tableView: UITableView, estimatedHeightForHeaderInSection section: Int) -> CGFloat {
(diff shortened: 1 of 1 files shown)


jeffaburt:
Hmm, there seem to be some issues with the example app and this code. Going to close this PR while I investigate.

image

jrmsklar:
@jeffaburt was that issue you pointed out ever resolved?

jeffaburt:
@jrmsklar https://github.com/xmartlabs/Eureka/pull/1257

jrmsklar:
Got it, thanks @jeffaburt!

PR #1257, opened 2017-09-29
Remove blank section headers/footers from plain tables on iOS 11

jeffaburt:
In iOS 11, table views now use estimated heights by default, which also means that cells and section header/footer views now self-size by default. The default value of the estimatedRowHeight, estimatedSectionHeaderHeight, and estimatedSectionFooterHeight properties is now UITableViewAutomaticDimension.

This is problematic when `tableView.style` is set to `.plain`. If there is no header or footer but we return `UITableViewAutomaticDimension` as the height, then UIKit will add space for a header or footer. 

This PR updates those methods to return `0` instead, that way no space is added.

## Screenshots
slice

*everything still looks good on a 4s running iOS 8.2*
image


jeffaburt:
Bump

jeffaburt:
@mtnbarreto do you mind checking this out when you get a free second?

mtnbarreto:
@jeffaburt Looks ok for me! Thanks for contributing!

jeffaburt:
Thanks for the merge!

PR #1278, opened 2017-10-10
Minor sample app UI glitch fix, fixed typo and allow having a value displayed for ButtonRow

danurna:
Hello there! Basically two things happen in this pull request.

### Fix of UI glitch in sample app
The `sliderRow` inside the sample app's `RowsExampleViewController` doesn't fit into the fixed height of 44pt [0]. I've changed the table view's row height to automatic.  

### Allow having a value displayed for `ButtonRow`
I've changed the `cellStyle` of `ButtonRow` to `.value1` in order to have an effect when using `displayValueFor`.

I am open for any feedback regarding this pull request. ✌️ 

[0] Screenshot 
eureka slider row glitch


Source/Rows/ButtonRow.swift
```diff
@@ -61,7 +61,7 @@ open class _ButtonRowOf<T: Equatable> : Row<ButtonCellOf<T>> {
     required public init(tag: String?) {
         super.init(tag: tag)
         displayValueFor = nil
-        cellStyle = .default
+        cellStyle = .value1
```
mats-claassen: The ButtonRow should have `.default` as its default cellStyle. It can always be changed in the row initializer:

```swift
ButtonRow() { row in
    row.cellStyle = .value1
}
```

mtnbarreto:
Seems that all was already fixed in master branch.

PR #1296, opened 2017-10-24
Fix SliderRow Layout problem

### LINKED ISSUE #324: SliderRow layout problem on iPhone5s/4s
Hi guys,

Great job on Eureka! I love it. 

Looks like there is a layout issue with the Slider row on smaller screen sizes. See the screenshot for example:
screen shot 2016-03-09 at 18 38 58

akabab:
Changes the SliderRow's layout constraints to fit a 44pt cell, correctly aligned & centred. 
Plus: valueLabel is now visible even if no title provided.

| Before  | After |
| --- | --- |
|  |  |

This fixes #324


mtnbarreto:
Hi @akabab , It would be great if we have the ability to show or not the value label. Could you implement it? 

> sometimes we want just to show the slider bar since the value doesn't matter or is shown in another view.

cerupcat:
This only appears to fix this by putting the slider and text on the same line. What if we want the current solution (where the slider is under), but also fix the current bug?

m-revetria:
@cerupcat, please check this https://github.com/xmartlabs/Eureka/issues/1316#issuecomment-356701867

LukA888:
Any update for SliderRow's UI Bug?

akabab:
@mtnbarreto Hey, sorry for the late reply. I've done small adds to show/hide the value label

mtnbarreto:
Hi @akabab ! thanks for contributing!

PR #1299, opened 2017-10-27
Added SelectionListRow and SelectorInlineRow

loromits:
Hi!  
Just wrote two custom rows and found them pretty generic, so wanted to share with everybody:
### SelectionListRow
SelectionListRow
Row with tableView and optional header(if title is present).
Supports single selection, customizable textLabel text color and table insets
### SelectorInlineRow
SelectorInlineRow
Inline row based on SelectionListRow. Collapses on value picking.


ilyapuchka:
Maybe https://github.com/EurekaCommunity will be a better place for that?

cleg:
@ilyapuchka I think this is better to be merged in main repository, as this rows look more universal and generic then those in Community. IMHO they're better then inline picker row

mtnbarreto:
Hey @loromits !! 
Thanks for contributing! I liked the rows! 
I think EurekaCommunity is the proper place to share these kind of rows. Do you have time to do so?

loromits:
Hey, ok I don't mind moving it to Eureka Community

mtnbarreto:
Hi @loromits I already gave you access to EurekaCommunity.  Looking forward to trying the new row types soon.

PR #1310, opened 2017-11-04
Call super.updateConstraints in SegmentedCell

pilot34:
Without that, we have crash for `SegmentedCell` with xib. Description of crash tells that we didn't call `super.updateConstraints()`


mtnbarreto:
Hi @pilot34 ! thanks for fixing the issue!!

PR #1333, opened 2017-11-24
Fix MultipleSelectorViewController Section Sorting

tjboneman:
- Corrected issue in section sorting function of MultipleSelectorViewController, where all options were placed in one section, with a section title based on the first option


mtnbarreto:
Thanks for contributing @tjboneman!!

PR #1339, opened 2017-12-02
Make default table header height and table footer height configurable

niklassaers:
I needed to be able to set table section header height or footer section height to 0 in some instances


mtnbarreto:
I'm ok with the ability to change it but at the same time i would like to refactor the FormViewController configuration properties to encapsulate them into a specific type. 
Eventually Eureka will not requite we extend FormViewController as we do nowadays.

PR #1344, opened 2017-12-07
Fix FieldCell titleLabel layout can overlap to its value label

banjun:
In addition to #965 ...

I found that the force updating constraints introduced in #965 is useful also when `titlePercentage` is not customized.

Without forcing that, title label can overlap to its value label.
The issue reproduces when iOS perform re-layout titleLabel (with the fact that titleLabel refs to `UITableViewCell.textLabel`) without updating its constraints. This case may occur in swiping, showing alert or actionsheet from the same section.

In the Eureka Example, adding ActionSheetRow after TextRow and set a long title for the TextRow reproduce the issue on iOS 11 (including iOS 11.2).

screen shot 2017-12-07 at 11 28 42

screen shot 2017-12-07 at 11 28 46

### DIFF
--- a/Source/Rows/Common/FieldRow.swift
+++ b/Source/Rows/Common/FieldRow.swift
@@ -403,6 +403,13 @@ open class _FieldCell<T> : Cell<T>, UITextFieldDelegate, TextFieldCell where T: 
 	open override func layoutSubviews() {
 		super.layoutSubviews()
 		guard let row = (row as? FieldRowConformance) else { return }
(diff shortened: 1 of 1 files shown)


banjun:
@quiekedei Do you have any idea for this?

mtnbarreto:
Hi @banjun! I was able to reproduce the issue following your indications. Thanks for contributing!!

banjun:
Thank you for review & merge!

PR #1349, opened 2017-12-12
Crash prevent Section.swift

pablojotaguardiola:
It's crashing when 'index' is greater than number of sections of tableView.

### DIFF
--- a/Source/Core/Section.swift
+++ b/Source/Core/Section.swift
@@ -44,7 +44,7 @@ extension Section : Hidable, SectionDelegate {}
 extension Section {
 
     public func reload(with rowAnimation: UITableViewRowAnimation = .none) {
-        guard let tableView = (form?.delegate as? FormViewController)?.tableView, let index = index else { return }
+        guard let tableView = (form?.delegate as? FormViewController)?.tableView, let index = index, index < tableView.numberOfSections else { return }
         tableView.reloadSections(IndexSet(integer: index), with: rowAnimation)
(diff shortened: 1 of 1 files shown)


mtnbarreto:
Hi @pablojotaguardiola. Seems to me you may have some issue that make it crash, number of 
table view sections must match the number of sections in Eureka Form.

From where are you calling section.road?

pablojotaguardiola:
This is not the code I was using, but you can see the problem is reloading section just after form building, in the viewDidLoad, for example:

screen shot 2017-12-18 at 21 59 00

pablojotaguardiola:
Sorry, @mtnbarreto, I didn't mention you.

mtnbarreto:
The issue happens due to that the Form UITableView is not on screen yet. Reload methods can only be used if the table view is on screen.

pablojotaguardiola:
@mtnbarreto I undestand, but I'm working in a big project, with 30 people. The project was working great with Swift 3, but at the moment we migrated to Swift 4, that crash showed up. 
If you are not including the commit soon, please just let me know, because I'd have to change every ViewController that uses Eureka forms to Fastlane works and we can upload a new version through pipelines.
I thought it was a good idea just to prevent unnecessary crashes.

mtnbarreto:
I'm not against merging the workaround. Anyway I recommend you to remove these unnecessary reloads() invocations shortly. I don't know its side effect.

PR #1350, opened 2017-12-12
fixed: NSForegroundColorAttributeName syntax changed to use in swift 3+

tegui:
Single line modified to avoid swift 3 compilations errors

NSAttributedStringKey.foregroundColor is changed for NSForegroundColorAttributeName


mtnbarreto:
Hi @tegui! you should use Swift 4 to compile Eureka

PR #1378, opened 2018-01-08
Fixes example about getting rows by tag

### LINKED ISSUE #1374: Incorrect example in Readme with Row<Date> type
In the Readme currently, in the FAQ section, "How to get a Row using its tag value" there is the following line:

    let dateRow2: Row<Date>? = form.rowBy(tag: "dateRowTag")

But if you copy that into a `FormViewController` subclass, you get the following compiler error:

> 'Row' requires that 'Date' inherit from 'BaseCell'

Not quite sure what was intended here, or perhaps it is just outdated?

Possibly supposed to say `Row<DateCell>` although would still need an `as? Row<DateCell>` at the end of the line then.

m-revetria:
Fixes #1374

### DIFF
--- a/README.md
+++ b/README.md
@@ -1023,7 +1023,7 @@ For instance:
 let dateRow : DateRow? = form.rowBy(tag: "dateRowTag")
 let labelRow: LabelRow? = form.rowBy(tag: "labelRowTag")
 
-let dateRow2: Row<DateCell>? = form.rowBy(tag: "Rows") as? Row<DateCell>
+let dateRow2: Row<DateCell>? = form.rowBy(tag: "dateRowTag") as? Row<DateCell>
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1023,7 +1023,7 @@ For instance:
 let dateRow : DateRow? = form.rowBy(tag: "dateRowTag")
 let labelRow: LabelRow? = form.rowBy(tag: "labelRowTag")
 
-let dateRow2: Row<Date>? = form.rowBy(tag: "dateRowTag")
+let dateRow2: Row<DateCell>? = form.rowBy(tag: "dateRowTag") as? Row<DateCell>
```
mtnbarreto: @m-revetria @mcousillas6  According Swift type inference we should not need need `as? Row<DateCell>` here.

PR #1379, opened 2018-01-10
Enable customising textfield color

AmirKamali:
This is a small change, but allows it to control over textfield color when not disabled.


mtnbarreto:
Hi @AmirKamali ! by doing this, one it the coiler get gray it will remain gray. 
Have you tested it?

mtnbarreto:
In order to customize enabled, disable text color you can do something like this...

```swift
URLRow.defaultCellUpdate = { cell, row in cell.textField.textColor = row.isDisabled ? .disabledColor : .enabledColor }
```

PR #1411, opened 2018-02-07
Eureka Forms: Temporarily caching deleted rows, so they're available for UITableView updates.

phomelvig:
Under certain conditions, the rows were removed from the kvoWrapper, and then UITableView was needing the row to animate it's removal.  This PR temporarily caches the row for UITableView to use.


infiniteNIL:
Why was this PR closed?

PR #1439, opened 2018-02-26
Update cell when tintColor changes

karlvr:
So the textColor can be updated if it uses the tintColor, which a lot of cells do. It appears that the tintColor is currently used before the cell is added to the window, so if the tintColor has been customised on the window the cells aren’t updated.

Calling row.updateCell() seems like the best thing to call, as that should encompass all code (built-in and custom) that uses the tintColor.

### DIFF
--- a/Source/Core/Cell.swift
+++ b/Source/Core/Cell.swift
@@ -150,6 +150,12 @@ open class Cell<T>: BaseCell, TypedCellType where T: Equatable {
         return result
     }
 
+    open override func tintColorDidChange() {
+        super.tintColorDidChange()
+
+        row.updateCell()
+    }
+
     /// The untyped row associated to this cell.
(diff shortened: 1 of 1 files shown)


mtnbarreto:
Hi @karlvr ! do you know with is the behavior of iOS native cells? I would like Eureka reproduce the same behavior.

karlvr:
@mtnbarreto Yes, this is the behaviour of any view / control that is added to a table cell that uses the `tintColor`. It's not a behaviour of table cells though, it's a behaviour of controls that use the tintColor. They all implement the `tintColorDidChange` method so that the `tintColor` stays in sync. Because in Eureka we are using the `tintColor` directly (rather than dropping in a `UIButton` etc) we need to implement `tintColorDidChange` to keep the tint in sync.

Issue #1110, opened 2017-06-17
makeRowVisible: scrollToRow feature request for inlineRow


yurevich1:
It would be nice if you will implement state function where the table will scroll (top or bottom of the cell) after inlinerow is rolled out.

To implement I've wrote these changes. Please, add them into main branch.

1.  at [BaseRow.swift](https://github.com/xmartlabs/Eureka/blob/547b9b8f66bf4f1d8e875b17a43e55fee5e1c016/Source/Core/BaseRow.swift#L104) add this method 

```
/**
     scrolls to wanted part of view
     */
    open func scrollToTop() -> Bool {
        return false
    }
```

2. at [InlineRowType](https://github.com/xmartlabs/Eureka/blob/547b9b8f66bf4f1d8e875b17a43e55fee5e1c016/Source/Core/InlineRowType.swift#L88)

```
let stt = scrollToTop()
cell.formViewController()?.makeRowVisible(inline,stt)
```

3. at [Core](https://github.com/xmartlabs/Eureka/blob/074ad432e697650087276e01c67bbb255e3ab1f4/Source/Core/Core.swift#L1052) change into 
```
func makeRowVisible(_ row: BaseRow, _ scrollToTop:Bool){
```

4. at [Core](https://github.com/xmartlabs/Eureka/blob/074ad432e697650087276e01c67bbb255e3ab1f4/Source/Core/Core.swift#L1055) change into 
```
tableView.scrollToRow(at: indexPath, at: scrollToTop ? .top :.bottom, animated: true)
```

***Usage*** :
override this method somewhere in new class.

```
//scrollToTop
    open override func scrollToTop() -> Bool {
        return true
    }
```

Thanks.

m-revetria:
Hi @yurevich1, seems that you have it already working. Would be great if you open a PR with the changes you've made so far. We'll review it and merge into main branch if everything looks good. 

Thanks!

yurevich1:
I will do this asap

yurevich1:
I can't push the chages. I'm receiving this.
```
git push --set-upstream origin final-scroll-position-type
Username for 'https://github.com': yurevich1
Password for 'https://yurevich1@github.com': 
remote: Permission to xmartlabs/Eureka.git denied to yurevich1.
fatal: unable to access 'https://github.com/xmartlabs/Eureka/': The requested URL returned error: 403
```

yurevich1:
@m-revetria please check my pull request

[...]

PR #1452, opened 2018-03-12
mod: Implement the final position of the cell after scrolling

yurevich1:
This feature can be overwritten by any child Eureka Cell


yurevich1:
https://github.com/xmartlabs/Eureka/issues/1110
I've recreated new pull request for a small feature. Please check and approve

yurevich1:
@mtnbarreto please, review this request

Source/Core/BaseRow.swift
```diff
@@ -127,6 +127,13 @@ open class BaseRow: BaseRowType {
 
     open func prepare(for segue: UIStoryboardSegue) {}
 
+    /**
+     Helps to pick destination part of the cell after scrolling
+     */
+    open var isScrollToTop: Bool {
```
mats-claassen: This is a compute property so that to change it you must subclass the Row. I think it would be better as a simple public variable.
Apart from that, as you open the possibility to change where the row should be scrolled to why not let the variable be a `UITableViewScrollPosition`?

I would say something like:
`public var inlineRowScrollPosition: UITableViewScrollPosition = .bottom`

And for one line documentation you should use `///` if possible to shorten it a bit.

What do you think?

yurevich1:
@mats-claassen Hi. I've changed the code regarding your comments. Please check.

PR #1465, opened 2018-03-16
Make _ActionSheetRow open

plummer:
Perhaps I'm wrong but this was the most convenient place to subclass for me, could we make this class open?


plummer:
Closing to clean up and request again later.

PR #1468, opened 2018-03-19
Implements support for UITableViewCellStyleSubtitle on FieldRow

txaiwieser:
Hey guys! First of all I would like to thank you all for the work you do! :D

**Case:**
Almost all row types in Eureka use a stock UIKit UITableViewCell as underlying cell. So if you change the UITableViewCellStyle of that cell if automatically changes the UI of the Eureka row.

But that are some rows that are backed by a custom Eureka row (FieldRow). So setting the UITableViewCellStyle of that cell doesn't affect the UI.

**Solution:**
I've changed the customConstraints method on FieldRow to change its layout if the row have a .subtitle style.

I develop in a way it doesn't modify the previous behaviour (keeping the same code). And just add an case for this new style.

The new layout code is very similar to the old one, so I guess its self explanatory.

We are using in production for some time. And we think this should be the available to everyone.

Any improvements or comments are welcome. Maybe we can improve it even more!

Thanks

### DIFF
--- a/Source/Rows/Common/FieldRow.swift
+++ b/Source/Rows/Common/FieldRow.swift
@@ -213,11 +213,17 @@ open class _FieldCell<T> : Cell<T>, UITextFieldDelegate, TextFieldCell where T: 
 
         if !awakeFromNibCalled {
             if let title = row.title {
-                textField.textAlignment = title.isEmpty ? .left : .right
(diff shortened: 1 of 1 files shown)


mats-claassen:
Thanks for the contribution!

txaiwieser:
🚀

Issue #1466, opened 2018-03-19
Add Chinese readme


Lebron1992:
Hi! I'm an iOS developer form China, and work for [Hacknocraft](http://hacknocraft.com/) in New York.

`Eureka` is elegant iOS form builder in Swift. Thanks for your contributions.

I want to add a Chinese readme for developers in China to help them use the library more comfortably. I would be glad to do this.

Lebron1992:
Ok, It’s my pleasure. I will create a PR when I’ve finished it.

Lebron1992:
@mats-claassen Please look at the PR: https://github.com/xmartlabs/Eureka/pull/1487

PR #1487, opened 2018-03-29
add Chinese readme

Lebron1992:
@mats-claassen Added Chinese readme. https://github.com/xmartlabs/Eureka/issues/1466


mats-claassen:
Thanks @Lebron1992!

Issue #41, opened 2015-10-18
Readonly mode as opposed to disabled


tp:
What do you think of a read-only mode that does not change the appearance to light grey, but does not allow editing either?

The use-case I have in mind is a global "Edit" button, that just enables editing on the form without it changing its appearance.

mtnbarreto:
@tp 
it's not a common scenario. Let's see if someone else ask for it to define if it worth adding it.
Anyway I think you can easily accomplish it by doing:

```swift
        for row in form.rows {
            row.baseCell.userInteractionEnabled = false
        }
```

tp:
Thanks this solves the issue nicely and looks exactly as I would expect from a "read-only" mode. :-)

Isuru-Nanayakkara:
I too would say this is a 'nice to have' feature. I know it's not too dofficult to do the workaround but still it would _look_ clean if I didn't have to call the `baseCell`.

davidrft:
I also think this is a nice to have future. I think this workaround:
`        for row in form.rows {
            row.baseCell.userInteractionEnabled = false
        }`
is not so good because it disables all kinds of interactions with the row, including scrolling, which is a interaction that a readonly mode should be able to handle.

radvansky-tomas:
Not really because if you make row with user interaction == false, then events are not fired, and therefore it cant react to selections...so its not a solution

radvansky-tomas:
So can you extend read-only to fields as well ?

gudleyd:
> Not really because if you make row with user interaction == false, then events are not fired, and therefore it cant react to selections...so its not a solution

Hi, did u solve this problem?

radvansky-tomas:
I've done some PR, wasnt accepted https://github.com/xmartlabs/Eureka/pull/1653

itimotin:
I did in this way
`form +++ Section("DATA")
             <<< TextRow(){
                $0.title = "Code"
                $0.tag = "Code"
                $0.cellSetup { cell, row in
                    cell.textField.isUserInteractionEnabled = false
                }
                }.onCellSelection{cell, row in
                    let iccCodeSectionViewController = CodesTableViewController()
                    self.navigationController?.pushViewController(iccCodeSectionViewController, animated: true)
}
`

[...]

PR #1489, opened 2018-03-29
Implement readOnly textAreaRow feature

davidrft:
Based on this [issue](https://github.com/xmartlabs/Eureka/issues/41) I added the possibility for a scrollable read only textAreaRow. I also changed the Example app to show its implementation. Hope it helps!

### DIFF
--- a/Source/Rows/TextAreaRow.swift
+++ b/Source/Rows/TextAreaRow.swift
@@ -1,4 +1,4 @@
-//  AlertRow.swift
+//  TextAreaRow.swift
 //  Eureka ( https://github.com/xmartlabs/Eureka )
 //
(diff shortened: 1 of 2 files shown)


Source/Rows/TextAreaRow.swift
```diff
@@ -226,6 +231,9 @@ open class _TextAreaCell<T> : Cell<T>, UITextViewDelegate, AreaCell where T: Equ
     }
 
     open func textViewShouldBeginEditing(_ textView: UITextView) -> Bool {
+        if let textAreaRow = self.row as? TextAreaRow, textAreaRow.textAreaMode == .readOnly {
```
mats-claassen: You should cast as `_TextAreaRow` here.

PR #1500, opened 2018-04-04
add 'selectableRowSetup' callback to SelectorViewController & MultipleSelectorViewController

tjboneman:
Currently the functionality to specify a custom `CellProvider` or `cellStyle` to be used for a `SelectorViewController`s rows is broken.

A `CellProvider` / `cellStyle` may be set on the row using the `selectableRowCellSetup`'callback, however this does not work as this callback occurs *after the cell has already been created.*

To resolve this issue, another callback `selectableRowSetup` is proposed, that is fired after the row is created, but before the cell is created.


mats-claassen:
Looks good. Could you also add a comment in the Readme in the [corresponding section](https://github.com/xmartlabs/Eureka#how-to-customize-selector-and-multipleselector-option-cells)

PR #1505, opened 2018-04-08
Support attributed strings in picker rows

anlaital:
This adds support for specifying titles for picker rows as attributed strings, allowing for further customisation of how the picker displays its titles. If `onProvideStringAttributes ` is not set, then previous functionality is retained.

### DIFF
--- a/Source/Rows/PickerRow.swift
+++ b/Source/Rows/PickerRow.swift
@@ -75,7 +75,7 @@ open class PickerCell<T> : Cell<T>, CellType, UIPickerViewDataSource, UIPickerVi
     open func numberOfComponents(in pickerView: UIPickerView) -> Int {
         return 1
     }
-    
+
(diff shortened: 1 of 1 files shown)


Source/Rows/PickerRow.swift
```diff
@@ -75,7 +75,7 @@ open class PickerCell<T> : Cell<T>, CellType, UIPickerViewDataSource, UIPickerVi
     open func numberOfComponents(in pickerView: UIPickerView) -> Int {
         return 1
     }
-
+
```
mats-claassen: Please remove this tab

anlaital: That should be a newline, or am I missing something? 😅

anlaital:
This should now be fixed.

mtnbarreto:
Closing due to lack of response.

PR #1539, opened 2018-05-08
added ability to change vertical padding of cell

uknowmeright:
now users can change the vertical padding in a FieldCell simply by calling:

cell.verticalPadding = 5

### DIFF
--- a/Source/Core/Cell.swift
+++ b/Source/Core/Cell.swift
@@ -26,6 +26,9 @@ import Foundation
 
 /// Base class for the Eureka cells
 open class BaseCell: UITableViewCell, BaseCellType {
+    
+    //for changing vertical padding of cell
+    open var verticalPadding: Int = 4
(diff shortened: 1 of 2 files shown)


Source/Core/Cell.swift
```diff
@@ -26,6 +26,9 @@ import Foundation
 
 /// Base class for the Eureka cells
 open class BaseCell: UITableViewCell, BaseCellType {
+    
+    //for changing vertical padding of cell
+    open var verticalPadding: Int = 4
```
mats-claassen: This should not be in BaseCell if you only use it in `_FieldCell`

mats-claassen: Also please keep the functionality as it is currently (i.e. 11)

mtnbarreto:
Closing due to lack of response.

Issue #1135, opened 2017-07-07
MultivaluedSection showInsertIconInAddButton property does not have any effect


SkylerSeamans:
Setting showInsertIconInAddButton to false in a MultivaluedSection still shows the green Insert button.

This can be observed in the Example project in ViewController.swift line 1625 (Multivalued Only Insert)

Environment: Eureka 3.0.0, Xcode 8.3.3 and iOS 10.3

harri121:
You can override the tableView delegate as a workaround:

`override func tableView(_ tableView: UITableView, editingStyleForRowAt indexPath: IndexPath) -> UITableViewCellEditingStyle {
        return .none
    }`

rafalwojcik:
@mtnbarreto Fixed in my PR :)

mtnbarreto:
#1557 fixes it.

PR #1557, opened 2018-05-24
Fix problem with property showInsertIconInAddButton

rafalwojcik:
Fix problem with property showInsertIconInAddButton don't hide plus button in MultivaluedSection

Problem mentioned in  #1135

### DIFF
--- a/Source/Core/Core.swift
+++ b/Source/Core/Core.swift
@@ -913,7 +913,7 @@ extension FormViewController : UITableViewDelegate {
             return .none
         }
         if section.multivaluedOptions.contains(.Insert) && indexPath.row == section.count - 1 {
-            return .insert
+            return section.showInsertIconInAddButton ? .insert : .none
         }
         if section.multivaluedOptions.contains(.Delete) {
             return .delete


Source/Core/Core.swift
```diff
@@ -912,7 +912,7 @@ extension FormViewController : UITableViewDelegate {
 			}
             return .none
         }
-        if section.multivaluedOptions.contains(.Insert) && indexPath.row == section.count - 1 {
+        if section.multivaluedOptions.contains(.Insert) && indexPath.row == section.count - 1 && section.showInsertIconInAddButton {
             return .insert
         }
```
mtnbarreto: I think the proper code is: 
```swift
 if section.multivaluedOptions.contains(.Insert) && indexPath.row == section.count - 1{ 
      return section.showInsertIconInAddButton ? .insert : .none
}
```

rafalwojcik: It will fall through so my code is also proper one. I don't see any advantages compared to your code.

mtnbarreto: For the last row if showInsertIconInAddButton is false and multivalued section allows Insertion and deletion your code display a delete icon when it should display nothing. 

Hope this clarify something.

rafalwojcik: Yes, now I understand. Let me change this.

rafalwojcik:
@mtnbarreto Changed :)

PR #1595, opened 2018-07-08
Add ability to customise the text color of UIPickerView used by PickerRow

AndrewBennet:
It would be very helpful to be able to customise the display of the UIPickerView used in PickerRows. Setting the background color is easy, but changing the text color is tricky. I came across the `UIPickerViewDelegate` function [`pickerView(_:attributedTitleForRow:forComponent:)`](https://developer.apple.com/documentation/uikit/uipickerviewdelegate/1614375-pickerview) which returns an `NSAttributedString` (which can specify color).

In my project which uses Eureka, I couldn't find any way that I could implement that function and reuse the existing PickerCell: I ended up having to duplicate the PickerCell / PickerRow code, and add the extra method to PickerCell. It would be very nice to have this customisation built-in, which is what this commit does.

I am using it in `defaultCellUpdate`; something like the following:

```
PickerInlineRow<Language>.InlineRow.defaultCellUpdate = { cell, _ in
   cell.pickerTextColor = .white
   cell.backgroundColor = .black
}
```

Note that the documentation states that:

> if your implementation of this method returns nil, the picker view falls back to using the string returned by the pickerView(_:titleForRow:forComponent:) method.

So cells with unspecified `pickerTextColor` will exhibit the existing behaviour.

### DIFF
--- a/Source/Rows/PickerRow.swift
+++ b/Source/Rows/PickerRow.swift
@@ -69,6 +69,8 @@ open class _PickerCell<T> : Cell<T>, CellType, UIPickerViewDataSource, UIPickerV
         picker?.dataSource = nil
     }
 
+    open var pickerTextColor: UIColor?
+
     open func numberOfComponents(in pickerView: UIPickerView) -> Int {
         return 1
     }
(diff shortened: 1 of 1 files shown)


AndrewBennet:
I've made a change to allow specification of general string attributes, rather than just the color

PR #1600, opened 2018-07-10
Update the documentation for creating custom inline rows

mats-claassen:


radvansky-tomas:
Can you please add simple example how to create inline row with custom cell via XIB ? 
As I tried using your documentation, but it does not work...

So documentation should state how to provide 2 cells in fact, as one is header and second will expand.

mats-claassen:
Have you read through our [blogposts](https://blog.xmartlabs.com/2016/09/06/Eureka-custom-row-tutorial/) on custom rows? That might be helpful. Creating a custom inline row should be similar to other custom rows.

The most important thing to not forget is to set the cellProvider in the init of the row:
`cellProvider = CellProvider<UserInfoCell>(nibName: "UserInfoCell")`

radvansky-tomas:
Thank you, I finally did it, and I will post complete solution too, to understand what I meant...its not clear in documentation, special the part about
pickers vs tables inside expanded row.

PR #1603, opened 2018-07-12
Update cells instead of reloading index paths and reselecting

enhorn:
Fixes issue: https://github.com/xmartlabs/Eureka/issues/1607

Now updates the cell instead of reloading the index path and reselecting it, to avoid unnecessary layout passes.

From
```swift
if !selectedIndexPaths.isEmpty {
    tableView.reloadRows(at: selectedIndexPaths, with: .none)
}
selectedIndexPaths.forEach {
    tableView.selectRow(at: $0, animated: false, scrollPosition: .none)
}
```

To
```swift
selectedIndexPaths.forEach {
    form[$0].updateCell()
}
```


enhorn:
@mats-claassen 
The issue is not the selection state, the issue is the reload.
In the linked issue, we have an embedded view controller, which never will be in any selection state.

mats-claassen:
Yes I got that. What I meant with the workaround is removing the row from `selectedIndexPaths` so that the reload won't be executed for that row. And then you would have to execute the `updateCell` in `viewWillAppear`.

mtnbarreto:
I would prefer to keep the code as it is and use @mats-claassen approach to skip entire table view row reload.

PR #1613, opened 2018-07-18
Add Section.prepend() method

btraas:
**Motivation**
`MultiValuedSection.append()` would append rows below the "add" action row. 
image
image

**Solution**
Add sister `prepend()` method to add to the top of the section.


mats-claassen:
Why not just use `section.insert(_, at:0)`? or at index `count-1`?

PR #1619, opened 2018-07-23
Add `insert(row: after:)` method which allows inserting rows after a …

### LINKED ISSUE #1617: Getting a hidden row index?
- Environment: Eureka, Xcode and iOS version you are using.

Eureka: 4.2.0
XCode: 9.4.1
iOS: 10-12

Use case:

I do load up a form, consisting of rows with tags A, B, C, D, ... up to J.
I do hide rows C, D and E, by setting `row.hidden = true` and `row.evaluateHidden()` for each.

Now I would like to insert a new row, whether hidden or not, after row D, which I can only reference by the tag.
If I get the row with `var rowD = form.rowBy(tag: "D")` that's ok, but I cannot get its index in the form.

`rowD.indexPath` is **nil**, because it is not visible and I cannot access the list of hidden rows.
`rowD.section?.index(of: rowD)` is also **nil**.
 
How to get the hidden row index?

mats-claassen:
…currently hidden row

Fixes #1617

### DIFF
--- a/Source/Core/Core.swift
+++ b/Source/Core/Core.swift
@@ -299,10 +299,12 @@ extension Condition : ExpressibleByStringLiteral {
 /**
 Errors thrown by Eureka
 
-- DuplicatedTag: When a section or row is inserted whose tag dows already exist
+ - duplicatedTag: When a section or row is inserted whose tag dows already exist
+ - rowNotInSection: When a row was expected to be in a Section, but is not.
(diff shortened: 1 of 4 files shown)


andraz-kozelj:
Hey guys, seems like a small change, bumping the issue since would come handy. Want to have main Eureka repo code, not my local fix.

PR #1634, opened 2018-08-13
Make `FieldCell's` textField public to allow overrides

mats-claassen:
It allows us to  use custom textfield classes in subclasses of FieldCell without having to duplicate the whole code.

Example for phone formatting using SHSPhoneTextField:

```swift
final class CustomPhoneCell : _FieldCell<String>, CellType {
    private lazy var phoneField : SHSPhoneTextField = {
        let textField = SHSPhoneTextField()
        textField.translatesAutoresizingMaskIntoConstraints = false
        textField.formatter.setDefaultOutputPattern("###-###-####")
        textField.textDidChangeBlock = { [weak self] txtField in
            self?.textFieldDidChange(txtField!)
        }
        return textField
    }()

    override var textField : UITextField! {
        get {
            return self.phoneField
        }
        set{
            self.textField = phoneField
        }
    }

    required init(style: UITableViewCellStyle, reuseIdentifier: String?) {
        for subView in contentView.subviews {
            if subView is UITextField {
                subView.removeFromSuperview()
            }
        }
        contentView.addSubview(phoneField)
    }

    required init?(coder aDecoder: NSCoder) {
        super.init(coder: aDecoder)
    }
}
```

### DIFF
--- a/Source/Rows/Common/FieldRow.swift
+++ b/Source/Rows/Common/FieldRow.swift
@@ -130,7 +130,7 @@ extension TextFieldCell {
 
 open class _FieldCell<T> : Cell<T>, UITextFieldDelegate, TextFieldCell where T: Equatable, T: InputTypeInitiable {
 
-    @IBOutlet public weak var textField: UITextField!
+    @IBOutlet open weak var textField: UITextField!
     @IBOutlet public weak var titleLabel: UILabel?
 
     fileprivate var observingTitleText = false


irtemed88:
> I think that concrete class should be responsible for setting `super.textField` and keep its reference ( either by adding the textField in the views hierarchy or having a string property.
> 
> Connecting the IBOutlet scenario should work too.

@mtnbarreto - Could you clarify how your recommendation would work when subclassing `_FieldCell` from outside of the Eureka module? I think since the scopes of `textField` and `titleLabel` are `public` they cannot be modified from outside the module. I think this also hides them from Interface Builder so you can't connect the outlets?

PR #1636, opened 2018-08-15
Add Int32Row and Int64Row

AndrewBennet:
I recently hit a snag in my app where I was trying to use an `IntRow` to allow input of a value which is to be stored in an `Int64`. Since `IntCell` is declared as `class IntCell: _FieldCell<Int>`, and the size of `Int` depends on the bitness of the platform, I have an issue when running on 32-bit devices (e.g. iPhone 5, iPad Mini).

It would be helpful to be able to use a row backed by 64-bit Int, regardless of the bitness of the device. To that end, I've added `Int32Row` and `Int64Row`.

### DIFF
--- a/Source/Rows/Common/FieldRow.swift
+++ b/Source/Rows/Common/FieldRow.swift
@@ -35,7 +35,16 @@ public protocol FieldRowConformance : FormatterConformance {
 }
 
 extension Int: InputTypeInitiable {
-
+    public init?(string stringValue: String) {
+        self.init(stringValue, radix: 10)
+    }
+}
+extension Int32: InputTypeInitiable {
+    public init?(string stringValue: String) {
+        self.init(stringValue, radix: 10)
+    }
+}
+extension Int64: InputTypeInitiable {
     public init?(string stringValue: String) {
         self.init(stringValue, radix: 10)
     }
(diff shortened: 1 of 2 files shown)


mats-claassen:
Hi @AndrewBennet I see that the cells are exactly the same for all three Int typed rows. Also the formatter is always the same. 

What about defining `_IntRow` generic for the cell part and `_IntCell` generic for the value. You could then create `Int64Cell`, `IntCell` and `Int32Cell` with only the required initializers. And you could create `Int64Row` simply as `_IntRow<Int64Cell>` and so on. The logic would not be repeated.

On the other hand I also think that it is very easy to add these kind of rows to any project that needs them and that they do not necessarily have to be in Eureka. You can simply add them to your project. And they are probably not used very often.

What do you think?

AndrewBennet:
Your suggested simplification sounds very sensible - I could definitely do that.

If you feel that these rows are unnecessary to include in Eureka, that's no problem, I'm happy to close the pull request. At the time, it felt like an limitation of Eureka - but you're right: it isn't very hard to add the custom rows to any project which requires them.

mats-claassen:
Our current intention is to not add too many rows (which have to be maintained) to Eureka but rather leave them as custom rows. I think this pull request can serve as reference if somebody wants to implement this feature.

Thanks for your understanding.

PR #1653, opened 2018-09-04
addButtonProvider disabled check

radvansky-tomas:
addButtonProvider might be disabled, but its action is propagated. And as multivaluedRowToInsertAt does not allow to pass "nil" value, we are not able to control flow for disabled buttons!


Source/Core/Section.swift
```diff
@@ -476,8 +476,11 @@ open class MultivaluedSection: Section {
     func initialize() {
         let addRow = addButtonProvider(self)
         addRow.onCellSelection { cell, row in
-            guard let tableView = cell.formViewController()?.tableView, let indexPath = row.indexPath else { return }
-            cell.formViewController()?.tableView(tableView, commit: .insert, forRowAt: indexPath)
+            if !row.isDisabled
```
mats-claassen: Can you leave this like `if !row.isDisabled {`, with the `{` in the same line, please?

radvansky-tomas: Done

Source/Core/Section.swift
```diff
@@ -42,37 +42,37 @@ public func == (lhs: Section, rhs: Section) -> Bool {
 extension Section : Hidable, SectionDelegate {}
 
 extension Section {
-
+
```
mats-claassen: Can you please remove all these empty spaces, here and below?

radvansky-tomas:
I don't know about what is this delay, but I've reopened that file and reformatted it, so hopefully you can read it know

mats-claassen:
What I meant with removing spaces is all the whitespaces you added in the commits. Empty lines should have no trailing whitespaces. You can set Xcode to do this in `Xcode -> Preferences -> Text Editing -> Automatically trim trailing whitespace`

radvansky-tomas:
I've opened that file and let xcode with settings you mentioned reformat that code. And results is just the same. I dont see any "extra" white lines there. Only usual formatting. If your formatter does something different, feel free to reformat it. Just accept this change, as is sitting here for ages.

mats-claassen:
It's not extra white lines but spaces. Normally Xcode does not automatically change it but it will delete the spaces when you edit the line.

In your case the easiest thing is to just discard all those changes using git. The changes should be only the lines of the changed code.

mtnbarreto:
Hi @radvansky-tomas I'm gonna commit the change so we do introduce these empty lines change to git history.

PR #1657, opened 2018-09-10
Remove sections from Form on deinit.

### LINKED ISSUE #1652: Memory leak after assign new value to form
I have a `FormViewController` instance that must show different forms base on category user has selected.

There is a function in my code that create a new instance of `Form` and append sections and rows to it the returns it: 
 
 
        
        fileprivate func generateForm(_ schema: Schema, postCreateVC: PostCreateViewController, category: PostCategory, isEditing: Bool) -> Form {
              let form = Form()
              //Append sections and rows to form property
              return form
        }

The first time user selects category, form will be generated and all the rows persistent count is correct in allocation instrument:
[https://imgur.com/wau18iJ](url)

 but the next times user selects other categories and I do something like:

`self.form = generateForm(...)`

Form will be replaced with previous form, but all rows persistent count will be added to previous numbers:
[https://imgur.com/K7r6l7N](url)

> Note that I tried `Form.removeAll()`, it removes current form references and persistent count doesn't add to previous one anymore, but it has a weird remove animation that I could not change it even with 
`override func deleteAnimation(forRows rows: [BaseRow]) -> UITableViewRowAnimation {}`

mats-claassen:
Fixes #1652

### DIFF
--- a/Source/Core/Form.swift
+++ b/Source/Core/Form.swift
@@ -60,6 +60,10 @@ public final class Form {
 
     public init() {}
 
+    deinit {
+        removeAll()
+    }
+
     /**
      Returns the row at the given indexPath
      */


mats-claassen:
Just noted this might provoke a crash if a Form is deallocated too early

PR #1666, opened 2018-09-23
Update Location Row after selection

leonmak:
Row does not update after selection.
Follow AlertRow to reload data after selection.


mats-claassen:
Wouldn't it be possible to just reload the row using `reloadRows(...)`?

mtnbarreto:
Hi Guys, once the user taps on the LocationRow, it should keep selected and Eureka reload it automatically when get back to the form from location selection. Somehow the cell does not get selected.
I would suggest to take a look why this is happening.

PR #1678, opened 2018-10-04
Adding `includeDisabled` to validate form method

txaiwieser:
This includes the possibility to validate e form without validating rows that are disabled. This is usefull for fields that are set by other fields input.


txaiwieser:
Important to say that this is a non-breaking change, since the parameter has a default value that doesn't change any previously behaviour

mtnbarreto:
@txaiwieser Thanks for contributing!

PR #1701, opened 2018-10-25
Avoid segmented row title label pushing value out of bounds

mats-claassen:
If you have a long title then set `cell.titleLabel?.numberOfLines = 0` and set an appropriate height.


mtnbarreto:
👍

PR #1723, opened 2018-11-19
Fix #1597: Cell disappears because of keyboard scrolling

### LINKED ISSUE #1597: Cell disappears because of keyboard scrolling
- Environment: Eureka 4.2.0, Xcode 9.3, iOS 11 (didn't try 10)

**TL;DR:** TextAreaRow disappears because of the keyboard.

**Full details:** I have a form with several fields and a few of them are not visible upon opening the form (because the form is too long).
When I highlight the first field, the keyboard appears and scrolls the table a bit making the next hidden cell to become visible (but under the keyboard). Well, this is the expectation. Because when the keyboard disappears, the next cell isn't visible and instead there is an empty space.
This behaviour is auto-fixed if the table reloads the missing cell.

Here is a video of the bug: https://youtu.be/RNqCsBSbb64

And here is the code of the missing cell:

```swift
override func viewDidLoad() {
    super.viewDidLoad()
    ...
    self.setupFormView()
}

private func setupFormView() {
    form.removeAll()
    let section1 = ...
    let section5 = self.setupDescriptionSection()
    ...
    form.append(contentsOf: [section1, section2, section3, section4, section5, section6])
}

private func setupDescriptionSection() -> Section {
    let section = Section()
    let rowDescription = TextAreaRow(RowIdentifiers.description) { [weak self] in
        $0.placeholder = L10n.eventHintDescription
        $0.textAreaHeight = TextAreaHeight.dynamic(initialTextViewHeight: 90)
        $0.value = self?.eventForm.eventDescription
        }.onChange { [weak self] in
            self?.eventForm.eventDescription = $0.value
    }
    section.append(rowDescription)
    return section
}
```

TopalSergey:


TopalSergey:
Fix #1597

PR #1731, opened 2018-12-04
Allow the row to reset its own value

BruunoMM:
Goal
===
The idea is to give the row the ability to reset its own value to a predefined **default**.

To do this, we create a new property to store the **default value**, that won't change along with it's **value**. Doing so, we allow the row to reset itself at anytime, since it knows its own default value.

The main idea was to create the possibility to do this in the whole `section` or `form`, just like setting the form to a default state. I couldn't find a way to do this as I didn't explore much. If you have any idea, feel free to suggest.

The main motivation was to make the code cleaner, and more understandable, as the row should be responsible for its values and the things related directly to it, in my opinion.

How to test
===
Run the `testSetDefaultValues` test.

### DIFF
--- a/Source/Core/Row.swift
+++ b/Source/Core/Row.swift
@@ -63,6 +63,13 @@ open class RowOf<T>: BaseRow where T: Equatable{
             return _value
         }
     }
+    
+    /// The default value of this row. Sets the value property when setted or when the setDefaultValue is called.
+    open var defaultValue: T? {
+        didSet {
+            value = defaultValue
+        }
+    }
 
     /// The untyped value of this row.
     public override var baseValue: Any? {
@@ -89,6 +96,11 @@ open class RowOf<T>: BaseRow where T: Equatable{
         validationErrors = rules.flatMap { $0.validateFn(value) }
         #endif
         return validationErrors
+    }
+    
+    /// Resets the value of the row. Settings it's value to it's default value.
+    public func resetValue() {
+        value = defaultValue
     }
 
     /// Add a Validation rule for the Row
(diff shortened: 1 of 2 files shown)


Source/Core/Row.swift
```diff
@@ -63,6 +63,13 @@ open class RowOf<T>: BaseRow where T: Equatable{
             return _value
         }
     }
+    
+    /// The default value of this row. Sets the value property when setted or when the setDefaultValue is called.
+    open var defaultValue: T? {
+        didSet {
+            value = defaultValue
```
mtnbarreto: I would prefer to avoid this value property set up when configuring `defaultValue`. Otherwise we would have some unexpected behavior. 
Please remove `didSet` and update the test accordingly. 
I guess we can also rename the property to `resetValue` to avoid any misinterpretation.

BruunoMM: The main idea was setting the default value / value once in the code. I mean, if I set the default value, that is the initial value for the row and if I don't want to use a default value then I can still set it through the value property. Do you really think it would lead to an unexpected behavior?

mtnbarreto: I guess we would run into setting the default value somewhere after setting the row's value and in this particular case in most of the time we will want to maintain the row `value` value. 

I guess the default value is super useful for reseting the row or the entire form by calling `form.reset()` that would be implemented.

BruunoMM: I understand. I have made the changes you asked. I renamed the method to `resetRowValue`, because it cannot have the same name of the property.
Do you have any idea on how to implement the `form.reset()` ?

mtnbarreto:
@BruunoMM Now it looks ok! Thanks for contributing!

Issue #1643, opened 2018-08-20
perform form.removeAll() throw Array out of bounds when have lot of Sections in Form 


hopy11:
When a lots of sections in a form (500+),perform form.removeAll() will throw array out of bounds at ：

```
extension Form: MutableCollection {

    // MARK: MutableCollectionType

    public subscript (_ position: Int) -> Section {
        get {
            //position equal kvoWrapper.section.count ,so throw out of bounds...
            return kvoWrapper.sections[position] as! Section
        }
}
```

But when there are few of sections in form,It's all right!!!
Why is that?thanks ;)

hopy11:
And I noticed when removed every single section it have to call hasBeenRemoveSection(),and in it,it will refresh tableView.
Can this result in wasted performance???
Is there have some batch method for form.removeAll()?? It will refresh tableView once!???

ialimz:
I have this issue too.
it causes about 400 crash on our app since last week

ialimz:
@mats-claassen in my case, I generate a new form after call previous form's `form.removeAll()`,
and in `heightForRowAtIndexPath` (in `Core.swift`) when it wants get the row by row[indexPath.section][indexPath.row], crash happens due to `kvoWrapper.sections.count` is 0 and not updated yet

ialimz:
@mats-claassen I overwrite `sectionsHaveBeenRemoved` in my FormViewController class with empty body to do nothing, it seems that problem fixed

ialimz:
Nope, my tests fails on reloadRows now

mats-claassen:
Have you tried just replacing the form of the FormViewController, without calling `removeAll`?

rlopez1990:
I have the same problem, does anyone know the best way to solve this problem without memory leaks?

mats-claassen:
Fixed in #1758

PR #1758, opened 2019-01-15
Change `removeAll` implementations to be faster

mats-claassen:
Having a form with 500 sections and removing all of them would take ~15 seconds because `removeAllObjects` of NSMutableArray was calling the observer for every section it removed. This way the observer is called only once.


rlopez1990:
Nice catch 👍, this pull request also fixes this issue: https://github.com/xmartlabs/Eureka/issues/1643, I've just checked it out.

PR #1784, opened 2019-03-04
Disable scroll on show inlineRow if position none 

antigp:
Some times it may be need to disable scroll on displaying inlineRow

UITableView.ScrollPosition.none -> scroll tableview to bottom of cell =(

### DIFF
--- a/Source/Core/Core.swift
+++ b/Source/Core/Core.swift
@@ -1109,6 +1109,7 @@ extension FormViewControllerProtocol {
     // MARK: Helpers
 
     func makeRowVisible(_ row: BaseRow, destinationScrollPosition: UITableView.ScrollPosition = .bottom) {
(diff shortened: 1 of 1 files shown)


mats-claassen:
Hi, @antigp I think this change is not wanted as there might be correct uses of `makeRowVisible` with a `.none` scroll position. 

What is your use case? You want to expand an inline row but not show it?

mtnbarreto:
@antigp Are you saying UIKit scrolls to bottom even though `destinationScrollPosition` is `.none`?

mats-claassen:
Hi @antigp I changed the behavior in #1788 as `ScrollPosition.none` does not mean no scrolling but minimum scrolling to get a row visible. Now you can accomplish your task by setting `destinationScrollPosition` to nil.

PR #1792, opened 2019-03-22
Fix backgroundColor for SwipeActions

### LINKED ISSUE #1790: Setting SwipeAction backgroundColor has no effect
I am using Eureka 4.3.1 in Xcode Version 10.1 (10B61).

Setting the `SwipeAction` backgroundColor does not work. It only uses the default grey color as you can see in the screenshot:

Simulator Screen Shot - iPhone XS - 2019-03-20 at 15 26 01

I verified the issue with the included example application. The relevant code is this part:

```swift
            <<< LabelRow("Actions Left & Right: iOS >= 11") {
                $0.title = $0.tag

                let moreAction = SwipeAction(style: .normal, title: "More") { (action, row, completionHandler) in
                    print("More")
                    completionHandler?(true)
                }

                let deleteAction = SwipeAction(style: .destructive, title: "Delete", handler: { (action, row, completionHandler) in
                    print("Delete")
                    completionHandler?(true)
                })

                $0.trailingSwipe.actions = [deleteAction,moreAction]
                $0.trailingSwipe.performsFirstActionWithFullSwipe = true

                if #available(iOS 11,*) {
                    let infoAction = SwipeAction(style: .normal, title: "Info", handler: { (action, row, completionHandler) in
                        print("Info")
                        completionHandler?(true)
                    })
                    infoAction.backgroundColor = .blue

                    $0.leadingSwipe.actions = [infoAction]
                    $0.leadingSwipe.performsFirstActionWithFullSwipe = true
                }
```

Expected: The last `infoAction` should have a blue background color.

Actual:  The action only has the default grey background color for normal actions.

mats-claassen:
ContextualAction's `backgroundColor` was set to a computed properties with an empty setter and a getter that always returned `nil`.

Now it is changed to `actionBackgroundColor` which gets/sets `backgroundColor` for `UITableViewRowAction` and `UIContextualAction`

Fixes #1790

### DIFF
--- a/Example/Example/Controllers/SwipeActionsController.swift
+++ b/Example/Example/Controllers/SwipeActionsController.swift
@@ -52,7 +52,7 @@ class SwipeActionsController: FormViewController {
                         print("Info")
                         completionHandler?(true)
                     })
-                    infoAction.backgroundColor = .blue
+                    infoAction.actionBackgroundColor = .blue
 
                     $0.leadingSwipe.actions = [infoAction]
                     $0.leadingSwipe.performsFirstActionWithFullSwipe = true
--- a/README.md
+++ b/README.md
@@ -619,7 +619,7 @@ let row = TextRow() {
                     //make sure you call the completionHandler once done.
                     completionHandler?(true)
                 })
-            infoAction.backgroundColor = .blue
+            infoAction.actionBackgroundColor = .blue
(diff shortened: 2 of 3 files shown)


funkenstrahlen:
Thank you 😊

zhkufish:
why still backgroundColor in release 5.0.0 version?

PR #1811, opened 2019-04-08
Correctly remove observers when rows are deleted

### LINKED ISSUE #1804: Removing rows/sections doesn't remove observers - Causes memory leak
- Environment: Eureka 4.3.1, Xcode 10.1,  iOS 12.1
- Edit: Also on Eureka 5.0.0, Xcode 10.2, iOS 12.1

After running form.removeAll to remove all rows/sections from our form then adding more rows we noticed a memory leak where rows weren't getting released properly.

After looking into it and stepping through the code it seems that form.removeRowObservers isn't removing observers properly, The below screenshots show the state of the rowObservers dictionary before and after the the code that should remove the "showFan2" observer. 

Screen Shot 2019-04-02 at 12 33 13 PM
Screen Shot 2019-04-02 at 12 33 30 PM

As far as I can tell this seems to be a case of editing the local copy arr without applying the changes to the actual observer array at form.rowObservers

I tried editing the eureka code to the below and that seemed to work for me. There might well be a better way to do it, just trying to show the issue.

```
func removeRowObservers(from taggable: Taggable, rowTags: [String], type: ConditionType) {
        for rowTag in rowTags {
            guard var arr = rowObservers[rowTag]?[type], let index = arr.index(where: { $0 === taggable }) else { continue }
            arr.remove(at: index)
            rowObservers[rowTag]?[type] = arr // + this line
        }
   }
```

mats-claassen:
Observers were not being removed correctly

Fix #1804

### DIFF
--- a/Source/Core/Form.swift
+++ b/Source/Core/Form.swift
@@ -330,8 +330,11 @@ extension Form {
 
     func removeRowObservers(from taggable: Taggable, rowTags: [String], type: ConditionType) {
         for rowTag in rowTags {
-            guard var arr = rowObservers[rowTag]?[type], let index = arr.firstIndex(where: { $0 === taggable }) else { continue }
-            arr.remove(at: index)
+            guard let arr = rowObservers[rowTag]?[type], let index = arr.firstIndex(where: { $0 === taggable }) else { continue }
+            rowObservers[rowTag]?[type]?.remove(at: index)
+            if rowObservers[rowTag]?[type]?.isEmpty == true {
+                rowObservers[rowTag] = nil
+            }
         }
     }
 


Source/Core/Form.swift
```diff
@@ -330,8 +330,11 @@ extension Form {
 
     func removeRowObservers(from taggable: Taggable, rowTags: [String], type: ConditionType) {
         for rowTag in rowTags {
-            guard var arr = rowObservers[rowTag]?[type], let index = arr.firstIndex(where: { $0 === taggable }) else { continue }
-            arr.remove(at: index)
+            guard let arr = rowObservers[rowTag]?[type], let index = arr.firstIndex(where: { $0 === taggable }) else { continue }
+            rowObservers[rowTag]?[type]?.remove(at: index)
+            if rowObservers[rowTag]?[type]?.isEmpty == true {
```
mtnbarreto: This can be written as `if rowObservers[rowTag]?[type]?.isEmpty ?? false {`

Issue #870, opened 2017-01-09
a row's isValid defaults to true, and I can't figure out how to bind a button `disabled` to the form`s is valid state


yoavschwartz:
Hi,

I am trying to adopt Eureka's new validation methods, but I'm having trouble since I want to conditionally enable and disable a button row based on if the form is valid or not. The problems are
 
1. Each row has an initial isValid value of true (```return validationErrors.isEmpty```) so if I do this my button always starts enabled despite an empty form
2. I can use ```row.validate()``` but then all my validation errors start showing up despite an empty form.

Do you know of any solutions to the problem?

3 things I could imagine are:
1. like in web forms having a "clean" state so I can say something like ```row.isValid || row.isClean```  
2. would be really nice to have a function that evaluates the validity of a row, without side-effects  (i.e without setting self.validationErrors).
3. I appreciate the value of encapsulation as much as the next guy, but would it be totally crazy to have a public getter with a private setter for rules?

starr0stealer:
@yoavschwartz 

You can disable a button if the form validation has any errors with the below sample.
For managing "Clean State", for now you would have to wire this up manually, reacting to onChange events, and keeping a Bool flag.

```swift
<<< ButtonRow() {
    $0.title = "Submit"
    $0.disabled = Condition.function(["... Tags ..."]) { form in
        return form.validate().count != 0
    }
}
```

starr0stealer:
I also think we should have a "Clean State" system built into Eureka. I believe that Form Validation should be executed if asking if the form is valid, that shouldn't change. Checking if the form is clean would be enough to know it wasn't touched and don't need to execute validations.

I can take a look at implementing "Is Clean" soon as I finish my current project, and if no one else has started a PR for it.

starr0stealer:
@yoavschwartz 

Eureka happens to already have `wasChanged` on `BaseRow`

I put together this quick and "dirty" `isClean` func in an extension

```swift
extension Form {

    public func isClean() ->Bool {
        for row in rows {
            if row.wasChanged {
                return false
            }
        }
        return true
    }
}
```

Using it, prevents validation from being executed

```swift
<<< ButtonRow() {
    $0.title = "Submit"
    $0.disabled = Condition.function(["... Tags ..."]) { form in
        return form.isClean() || form.validate().count != 0
    }
}
```

yoavschwartz:
[...]

PR #1817, opened 2019-04-16
added quietly parameter for use with non-ui validation

dleute:
I did not read any contributing documents of any kind. I needed to solve a problem and solved it. Just decided to share the solution. I am open to providing tests/documentation, but wanted feedback first to see if this is an appropriate solution.

On #870 several suggestions were made to make it easier to tie a save button to validation results. I tried to solve this problem by extending Form and Row and BaseRow... etc. In the end I found it was too much private variables to do that way. So I went and added a "quietly" parameter to Form.validate and Row.validate.

Since this parameter defaults to false, I can't see how it would have any effect on normal usage.

I then use Form.validate(quietly: True) in onRowValidationChanged and in onChanged with a closure to update the ui save button.

This has the effect of frequently re-validating the entire form. I couldn't see a way around that in the current architecture. Since saving ValidationErrors is what triggers various callbacks, I simply prevented that from happening. Ideally, one would be able to simply read in all the current validation errors after doing a form validation on init. Once they change to 0, then the form is valid. And validation wouldn't have to be constantly re-done for the entire form.

This allowed "empty" forms and rows not to show validation errors until interacted with but still let the Save button state be 100% current.

I am opening two other issues I have discovered while writing this. A redux of Number/Decimal formatter and SegmentedRow not calling onRowValidationChanged.

### DIFF
--- a/Source/Core/Form.swift
+++ b/Source/Core/Form.swift
@@ -393,13 +393,17 @@ extension Form {
 extension Form {
 
     @discardableResult
-    public func validate(includeHidden: Bool = false, includeDisabled: Bool = true) -> [ValidationError] {
+    public func validate(includeHidden: Bool = false, includeDisabled: Bool = true, quietly: Bool = false) -> [ValidationError] {
         let rowsWithHiddenFilter = includeHidden ? allRows : rows
         let rowsWithDisabledFilter = includeDisabled ? rowsWithHiddenFilter : rowsWithHiddenFilter.filter { $0.isDisabled != true }
         
         return rowsWithDisabledFilter.reduce([ValidationError]()) { res, row in
             var res = res
-            res.append(contentsOf: row.validate())
+            if (quietly) {
+                res.append(contentsOf: row.validate(quietly: quietly))
+            } else {
+                res.append(contentsOf: row.validate())
+            }
             return res
         }
     }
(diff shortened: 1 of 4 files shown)


Source/Core/Form.swift
```diff
@@ -393,13 +393,17 @@ extension Form {
...
 
     @discardableResult
-    public func validate(includeHidden: Bool = false, includeDisabled: Bool = true) -> [ValidationError] {
+    public func validate(includeHidden: Bool = false, includeDisabled: Bool = true, quietly: Bool = false) -> [ValidationError] {
         let rowsWithHiddenFilter = includeHidden ? allRows : rows
         let rowsWithDisabledFilter = includeDisabled ? rowsWithHiddenFilter : rowsWithHiddenFilter.filter { $0.isDisabled != true }
         
         return rowsWithDisabledFilter.reduce([ValidationError]()) { res, row in
             var res = res
-            res.append(contentsOf: row.validate())
```
mats-claassen: You do not need the `if` here. You could just call `validate` with quietly.

dleute: Are you saying just do:

row.validate(quietly: quietly).

That would work.

dleute: Fixed

dleute:
Thanks!

PR #1833, opened 2019-05-14
Fix keyboard scrolling caret invisible

jonkan:
Fixed bug where the table view would scroll too far up when the keyboard shows, when presented in a tab bar controller with a translucent tab bar. 
If the tab bar is translucent the form's table view extends behind the tab bar to the bottom screen edge and since the content inset accounts for the tab bar, the `newBottomInset` becomes too large.

The bug becomes really annoying with e.g. a tall TextAreaRow's on an iPad in landscape, as you can't see what you type:
invisible-caret

### DIFF
--- a/Source/Core/Core.swift
+++ b/Source/Core/Core.swift
@@ -1012,7 +1012,10 @@ extension FormViewController {
         let endFrame = keyBoardInfo[UIResponder.keyboardFrameEndUserInfoKey] as! NSValue
 
         let keyBoardFrame = table.window!.convert(endFrame.cgRectValue, to: table.superview)
-        let newBottomInset = table.frame.origin.y + table.frame.size.height - keyBoardFrame.origin.y + rowKeyboardSpacing
+        var newBottomInset = table.frame.origin.y + table.frame.size.height - keyBoardFrame.origin.y + rowKeyboardSpacing
+        if let tabBar = tabBarController?.tabBar, tabBar.isTranslucent {
+            newBottomInset -= tabBar.frame.size.height
+        }
         var tableInsets = table.contentInset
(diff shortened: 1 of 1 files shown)


funkenstrahlen:
I see a similar issue as described here: https://github.com/xmartlabs/Eureka/issues/1952

However my code does not use a Tabbar. I took a look at your fix and it might work as you decrease the contentInset with your code. But I think it is just a coincidence that it works using the tabbar height. I feel like the correct adjustment for the offset needs to come from somewhere else.

funkenstrahlen:
@jonkan Are you also running this on iOS 13? I think I found the root cause of this issue and created a PR with a fix: https://github.com/xmartlabs/Eureka/pull/1953

jonkan:
> I see a similar issue as described here: #1952
> 
> However my code does not use a Tabbar. I took a look at your fix and it might work as you decrease the contentInset with your code. But I think it is just a coincidence that it works using the tabbar height. I feel like the correct adjustment for the offset needs to come from somewhere else.

I think the correct solution is to subtract the `tableView.safeAreaInset.bottom` as @yunhao did in #1960. My solution only fixed the case where the tableView's visible area was limited by a tab bar (where `tabBar.frame.size.height` equals `tableView.safeAreaInset.bottom`).

> @jonkan Are you also running this on iOS 13? I think I found the root cause of this issue and created a PR with a fix: #1953

Yes and no. We did encounter this bug before iOS 13 (iOS 12 I think) and then, this fix (and probably #1960) solved it. However now, on iOS 13, the bug is back in a slight variation. I'm guessing we're experiencing the same issues as you described in https://github.com/xmartlabs/Eureka/issues/1952

I'm closing this PR since #1960 has landed

Issue #1838, opened 2019-05-21
StepperRow Image / Text overlap


nidegen:
- Environment: Eureka 5.0.0, Xcode 10.2 and iOS 12.2 

Problem: When adding an image to the stepper row cell, the title overlaps the image.

Stringely, when the form updates with `evaluateHidden()` of another section, the title jumps to the right place, only to move back when using the stepper button..

Is that intentional? I can have a look for a fix if not.
```
<<< StepperRow() { row in
      row.title = "Title"
      row.displayValueFor = { value in
        guard let value = value else { return nil }
        return "\(Int(value))"
      }
    }.cellSetup { cell, row in
      cell.imageView?.image = UIImage(named: "image_name")
      }
```

mats-claassen:
Could you post a screenshot? However from what you say it does seem to be an issue. The text should never overlap with the image. Must be an unhandled case in the constraints. Fixes are welcome

nidegen:
![a](https://user-images.githubusercontent.com/8779423/58558139-551c6780-8220-11e9-8388-d15a9d588dea.png)
![t](https://user-images.githubusercontent.com/8779423/58558141-55b4fe00-8220-11e9-8a2f-bc80e3b1eca1.png)

This is how the two states look. Any tips or hints for the fix? Could it be a autolayout bug?

mats-claassen:
It seems the `StepperRow` has no constraints for the `imageView`.
It should be structured more like `SliderRow`, for example, having constraints set for different cases (like having an image or not) and updating them in `updateConstraints`

nidegen:
see #1857 1857

nidegen:
solved in #1857

PR #1857, opened 2019-06-03
Fix StepperRow layout when imageView is set

nidegen:
Addressing #1838 

I ended up copying the code from SliderRow and adapting it slightly.

One Remark:  I prefer the value label text color  to be grey like in the SliderRow, but I have left it as it was for now.

### DIFF
--- a/Source/Rows/StepperRow.swift
+++ b/Source/Rows/StepperRow.swift
@@ -8,18 +8,18 @@
 
 import UIKit
 
-/// The cell of the StepperRow
+// MARK: StepperCell
+
 open class StepperCell: Cell<Double>, CellType {
   
-    private var awakeFromNibCalled = false
   
-    @IBOutlet open weak var titleLabel: UILabel!
-    @IBOutlet open weak var valueLabel: UILabel!
     @IBOutlet open weak var stepper: UIStepper!
-  
-    open var formatter: NumberFormatter?
-  
-    public required init(style: UITableViewCell.CellStyle, reuseIdentifier: String?) {
+    @IBOutlet open weak var valueLabel: UILabel!
+    @IBOutlet open weak var titleLabel: UILabel!
+
+    private var awakeFromNibCalled = false
+
+    required public init(style: UITableViewCell.CellStyle, reuseIdentifier: String?) {
         super.init(style: .value1, reuseIdentifier: reuseIdentifier)
       
         NotificationCenter.default.addObserver(forName: UIContentSizeCategory.didChangeNotification, object: nil, queue: nil) { [weak self] _ in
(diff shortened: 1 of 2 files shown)


Source/Rows/StepperRow.swift
```diff
@@ -8,88 +8,154 @@
...
+    let vContraint = NSLayoutConstraint(item: stepper!, attribute: .centerY, relatedBy: .equal, toItem: contentView, attribute: .centerY, multiplier: 1, constant: 0)
+    dynamicConstraints.append(vContraint)
+    contentView.addConstraints(dynamicConstraints)
+  }
+  
 }
 
-// MARK: StepperRow
 
-open class _StepperRow: Row<StepperCell> {
```
mats-claassen: You should leave an open superclass to allow overrides

Source/Rows/StepperRow.swift
```diff
@@ -8,88 +8,154 @@
...
+      self.titleLabel = title
+      
+      let value = detailTextLabel
+      value?.translatesAutoresizingMaskIntoConstraints = false
+      value?.setContentHuggingPriority(UILayoutPriority(500), for: .horizontal)
+      value?.adjustsFontSizeToFitWidth = true
+      value?.minimumScaleFactor = 0.5
+      self.valueLabel = value
+      let font = self.valueLabel.font!
+      self.valueLabel.font = UIFont.monospacedDigitSystemFont(ofSize: font.pointSize, weight: font.weight)
```
mats-claassen: I don't think it is necessary to set this font here, and if then prefer `systemFont(ofSize:)`.

Source/Rows/StepperRow.swift
```diff
@@ -8,88 +8,154 @@
...
-        row.updateCell()
+    selectionStyle = .none
+    stepper.minimumValue = 0
+    stepper.maximumValue = 10
+    stepper.addTarget(self, action: #selector(StepperCell.valueChanged), for: .valueChanged)
+  }
+  
+  open override func update() {
+    super.update()
+    titleLabel.text = row.title
```
mats-claassen: This is done in `super.update`

Source/Rows/StepperRow.swift
nidegen: ok sure. As i was copying from SliderRow first, the numbers were positioned after the stepper. There, it shifted on ever digit change. There the monospaced digit font was a great fit.

Source/Rows/StepperRow.swift
nidegen: sure, that comes from SliderView - ok if I fix it there too then?

nidegen:
OK, I adressed everything and cleaned up the indentation. Didn't touch SliderRow though. ptal

Source/Rows/StepperRow.swift
mats-claassen: Yes

PR #1860, opened 2019-06-04
Set iOS deployment target to 9.0

PatrickNLT:
There is no obvious reason why we should restrict the deployment target to iOS 9.3 instead of 9.0, so I'm proposing this change.


mats-claassen:
@PatrickNLT do you have users using iOS version >= 9.0, < 9.3? As far as I know there is no restriction for any device that can update to iOS 9 to not update to 9.3.
On the other hand, Eureka does not use any newer features so we could set it to iOS 9.0

PatrickNLT:
@mats-claassen Yes, a very small part of our users are running iOS version >= 9.0, < 9.3. Since there is no harm in supporting those versions in our app, I don't see any reason why I should exclude them.

PR #1872, opened 2019-06-22
Fixed cuithemestore warning.

ykphuah:
There is a warning 
CUIThemeStore: No theme registered with id=0
When running my app compiled with Eureka, or the Examples app also.

The change is just to move the png files into Assets.xcassets folder instead of in Eureka.bundle


mats-claassen:
Did you try installing Eureka from this branch? Were the chevron assets there? Also, you did not remove the reference to the bundle from the podspec, nor did you include the Assets folder in the podspec. So it might still be using the bundle if you install it from this branch?

Issue #1871, opened 2019-06-21
What about adding Swift Package Manager support?


funkenstrahlen:
Xcode 11 allows adding swift packages very easily. Have you thought about adding SPM support to Eureka? I could not find any old discussion about this here, therefore I would like to start one!

I personally would love to use Eureka via swift package manager. It does not support assets though. However as far as I can tell this should not be an issue for Eureka.

mats-claassen:
It is good to have this discussion. It might be worth it if several people request it. I do not have an idea of how many projects are using SPM compared to Cocoapods or Carthage.

thebluepotato:
For now, since SPM support in Xcode and for iOS/tvOS/watchOS projects has only been added in Xcode 11 beta 1, there aren't many developers using SPM, but I guess more and more are trying to adopt it since it's easy to manage and keep up-to-date.

So in short: SPM is used way less than Carthage or Cocoapods _for now_, but I think it'll change soon!

thebluepotato:
Based on the Podspec, the main if not only obstacle is the Eureka.bundle file which can't be used with SwiftPM. I've written the `Package.swift` file in approximately 2 minutes, builds and tests perfectly. I just expect it to complain when the chevrons are used. PM available soon for discussion.

kostyan5:
+1 for wanting SPM support and agree with @thebluepotato that very soon SPM will take over Carthage and Cocoapods.

dan-hart:
I agree with @kostyan5, It's only a matter of time, and then SPM will be the dominant dependency manager.

Satwa:
SPM is way easier as it is integrated to Xcode, definitely needs the support for this lib

NathanFallet:
I’m already adding my libs with SPM so it would be great to add support for it. I don’t succeed using Carthage or CocoaPods...

txaiwieser:
Any news on this?

txaiwieser:
I saw that #1877 implements all the basic functionality, its just missing the bundle

DanTheColoradan:
I tried Swift Package Manager today. It looks pretty good and would use if for eureka (instead of Carthage). Did this get resolved?

nidegen:
I'd love to have it too. Couldn't we just draw the chevrons in the `.bundle` instead of using image files?

For me even a version without the `.bundle` would be ok:)

macblazer:
[...]

PR #1877, opened 2019-06-26
Adding support for SPM

thebluepotato:
Adds support for SPM, based on #1871 

CAUTION: the `Package.swift` file does not include a reference to `Eureka.bundle` because does not (yet) support ressources and assets (should change soon though AFAIK).


thebluepotato:
Regarding #1808 : this PR contains the test target for Eureka as well and I'm not sur Accio will stay for too long now that SPM supports iOS and others.

BalazsSzamody:
Hey, I found/created a solution for the bundle issue. For some reason I could't build it in the example project. Take a look at my branch: 
https://github.com/BalazsSzamody/Eureka/tree/SPM-Support

Let me know if I should make a PR.

AndrewBennet:
Just a note to say that I've got another pull request (https://github.com/xmartlabs/Eureka/pull/1976) with a different solution: rather than jump through hoops to get a bundle working, the images in the bundle can be drawn in about 35 lines of code instead.

mats-claassen:
This feature has been added in #1976

PR #1889, opened 2019-08-13
implemented dark mode

waterskier2007:


ykphuah:
> @mats-claassen can this be approved soon? iOS 13 is likely coming out in 2 weeks..

I think the issue is that xcode 11 have not gone into GM yet, without which developers can't build this and publish anyway.

funkenstrahlen:
I just tested this patch in my app and it works fine as expected.

Issue #1923, opened 2019-10-08
iOS 13 Swipe destructive doesn't work


ykphuah:
Can be reproduced with the sample app on iPhone 11 simulator running on iOS 13.1.

The Swipe Actions example doesn't remove the row in iOS 13, it worked perfectly in iOS 12.

Somehow in my app, there's another area which I use SwipeAction and it works well. The part that have issue, the actions actually gets triggered, just that visually the row is not removed. Workaround in my app is to back out and come back in again.

renatoamazillo:
I think I am seeing that issue here.  It happens if you have specified your custom action handler to (trailingSwipe.actions), and even calling `completionHandler!(true)` inside your custom handler, the rows doesn't go away, correct?

renatosc:
I think I am seeing that issue here. It happens if you have specified your custom action handler to (trailingSwipe.actions), and even calling completionHandler!(true) inside your custom handler, the rows doesn't go away, correct?

ykphuah:
Yeah. Some people think this is the right behavior. Then it is up to the handler to delete it. Maybe it can handle validation?

renatoamazillo:
I confirm that this behavior is due to a change from Apple in iOS 13, although it is not clear if it was a change made in purpose or is a bug that Apple may fix in a new iOS 13 update.   
Looking Eureka code, I also noticed that when you don't specify your custom action handler, Eureka is deleting the cell for you inside the [tableview commit editingStyle](https://github.com/xmartlabs/Eureka/blob/ad75e04120edcdf171a41dadb738a8ef0b92fc43/Source/Core/Core.swift#L842) delegate.

So, for now, a quick fix for this issue is just for you to manually remove the cell inside your custom action handler, something like this:

```
let deleteAction = SwipeAction(style: .destructive, title: "Delete") { (action, row, completionHandler) in
                                    if #available(iOS 13.0, *) { // as of iOS 13.2.2, completionHandler(true) is not removing the row, so we need to remove it manually
                                        row.section?.remove(at: row.indexPath!.row)
                                    }
                                    completionHandler!(true) // completes the deletion effect
                                }
```

renatosc:
I confirm that this behavior is due to a change from Apple in iOS 13, although it is not clear if it was a change made in purpose or is a bug that Apple may fix in a new iOS 13 update.
[...]

PR #1944, opened 2019-11-20
fix for iOS 13 that was not removing the row automatically anymore…

renatosc:
…even after calling the completion handler in a destructive swipe action.

This bug is easily reproducible using the Eureka Example Project: Running the app, just choose the "Swipe Actions" row and try to remove a row. If you are running iOS 13, the row will not be removed while it does get removed if the app is running a former iOS version.

This PR fixes the [Issue 1923](https://github.com/xmartlabs/Eureka/issues/1923)


renatosc:
Thanks @mats-claassen .

Issue #1952, opened 2019-12-12
ScrollView contentOffset not adjusted correctly when form is displayed as page sheet and keyboard is displayed


funkenstrahlen:
I discovered an issue with adjusting the form ScrollView `contentInset` and `contentOffset` when the user edits text in a `TextAreaRow`. This becomes a problem as soon as the form is presented as an **page sheet** on iOS 13. It looks like the scrolling does not consider the additional top space added by the new page sheet presentation style.

![2019-12-12 21-31-01 2019-12-12 21_31_58](https://user-images.githubusercontent.com/144173/70746835-e2d82400-1d26-11ea-86cc-0b09c6c486eb.gif)

This only happens when the view is displayed with the page sheet style (space at the top).

Also the behavior is not consistent as you can see in this clip:

![2019-12-12 21-37-07 2019-12-12 21_37_46](https://user-images.githubusercontent.com/144173/70747188-bc66b880-1d27-11ea-9be7-d2d080fe439f.gif)

The same code on iOS 12 without the page sheet presentation style: Also not consistent but does not scroll badly so the first line is invisible:

![2019-12-12 21-40-40 2019-12-12 21_41_07](https://user-images.githubusercontent.com/144173/70747425-4282ff00-1d28-11ea-865c-f8fea979227f.gif)

I guess this is the code that needs some update to fix this:
https://github.com/xmartlabs/Eureka/blob/29242596a2331c4e3be6a2481e8249861e878b6e/Source/Core/Core.swift#L1009-L1037

I tried to figure it out myself, but could not find a solution yet.

Environment: Eureka 5.1.0, Xcode 11.3

funkenstrahlen:
This might be related to this PR: https://github.com/xmartlabs/Eureka/pull/1833

However I do not use any Tabbar in my code.

funkenstrahlen:
Probably also related to https://github.com/xmartlabs/Eureka/issues/1938.

funkenstrahlen:
I did some more debugging and realized this issue is caused by `table.scrollToRow(at: selectedRow, at: .none, animated: animateScroll)`. It does not respect the additional `safeAreaInsets` on top added by the page sheet presentation.

Seems like an iOS 13 bug even though I am already running on iOS 13.3.

funkenstrahlen:
I found a workaround:

```swift
            if let selectedRow = table.indexPath(for: cell) {
                var rect = table.rectForRow(at: selectedRow)
[...]

PR #1953, opened 2019-12-13
Fix scrollToRow to respect safe area insets

funkenstrahlen:
There is a bug in iOS that causes `UITableView.scrollToRow` to not respect safe area insets. Therefore it does scroll to the wrong position.

A demo and description of the issue can be found here: https://github.com/xmartlabs/Eureka/issues/1952

This PR fixes the issue by using `scrollRectToVisible` instead and adding the safeAreaInset manually.


yunhao:
Thanks for the pull request. I have a similar issue. 
However, I don't think the change you make is correct. The behavior of `FormViewController` is unexptected because it totally ignores safe area when adjusting table view content inset. The `rect` is correct. We should fix `contentInset` instead of `rect`.

Issue #1959, opened 2019-12-25
FormViewController is incompatible with safe area when handling keyboard events


yunhao:
- Eureka 5.1.0
- Xcode 13.3
- iOS 13.3

The example project is not updated with safe area. The incompability not only makes the UI seems strange, but also conceal a bug in FormViewController when handling keyboard events. 

<img width="33.3%" alt="Simulator Screen Shot - iPhone 11 Pro - 2019-12-25 at 14 27 13" src="https://user-images.githubusercontent.com/10205466/71435136-a93dea80-2722-11ea-86a5-7a8274aa9b0e.png">

When safe area layout guide is enabled, `tableView` should fullfill it's parrent view. `tableView` will adjust it's contentInset automatically according to safe area. 

There is a new API after iOS 11:
> **adjustedContentInset**
The insets derived from the content insets and the safe area of the scroll view.

<img width="100" alt="Screen Shot 2019-12-25 at 14 38 08" src="https://user-images.githubusercontent.com/10205466/71435491-2f0e6580-2724-11ea-92eb-357ba7edecc3.png">

And then we will find an issue when tapping a TextRow. There are some extra spacing between keyboard and active row.

<img width="33.3%" src="https://user-images.githubusercontent.com/10205466/71437320-1dc95700-272c-11ea-8d04-a17a8caa467c.gif">

These codes solve the trouble:

```swift
open class FormViewController: UIViewController, FormViewControllerProtocol, FormDelegate {
    
    @objc open func keyboardWillShow(_ notification: Notification) {
       // ...
        var newBottomInset = table.frame.origin.y + table.frame.size.height - keyBoardFrame.origin.y + rowKeyboardSpacing
        if #available(iOSApplicationExtension 11.0, *) {
            newBottomInset = newBottomInset - view.safeAreaInsets.bottom
        }
        // ...
    }
}
```

yunhao:
There is an hacking without editing any source code.

```swift
class MyFormViewController: FormViewController {

    override func viewSafeAreaInsetsDidChange() {
        super.viewSafeAreaInsetsDidChange()
        rowKeyboardSpacing = -view.safeAreaInsets.bottom
    }

}
```

yunhao:
#1960

yunhao:
I create a pull request #1960

mats-claassen:
Fixed in #1960

PR #1960, opened 2019-12-26
Fix table view contentInset adjustment with safe area when keyboard shows

yunhao:
UITableView will adjust it's `contentInset` with safe area. The adjusted `contentInset` can be derived from `adjustedContentInset` api. When table view is influenced by safe area, `Eureka 5.1.0` has an issue that `contentInset` is adjusted too much after keyboard shows.

Solve #1959 #1953

### DIFF
--- a/Source/Core/Core.swift
+++ b/Source/Core/Core.swift
@@ -1012,7 +1012,10 @@ extension FormViewController {
         let endFrame = keyBoardInfo[UIResponder.keyboardFrameEndUserInfoKey] as! NSValue
 
         let keyBoardFrame = table.window!.convert(endFrame.cgRectValue, to: table.superview)
-        let newBottomInset = table.frame.origin.y + table.frame.size.height - keyBoardFrame.origin.y + rowKeyboardSpacing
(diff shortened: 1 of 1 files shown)


Source/Core/Core.swift
```diff
@@ -1012,7 +1012,10 @@ extension FormViewController {
         let endFrame = keyBoardInfo[UIResponder.keyboardFrameEndUserInfoKey] as! NSValue
 
         let keyBoardFrame = table.window!.convert(endFrame.cgRectValue, to: table.superview)
-        let newBottomInset = table.frame.origin.y + table.frame.size.height - keyBoardFrame.origin.y + rowKeyboardSpacing
+        var newBottomInset = table.frame.origin.y + table.frame.size.height - keyBoardFrame.origin.y + rowKeyboardSpacing
+        if #available(iOSApplicationExtension 11.0, *) {
```
mats-claassen: Why not `iOS` instead of `iOSApplicationExtension`?

yunhao: It's weird. `iOSApplicationExtension` is auto fixed by Xcode. Code has been updated with `#available(iOS 11.0, *)` now.

PR #1976, opened 2020-01-18
Remove bundle resources (reimplemented in code) to support Swift Package Manager

AndrewBennet:
The Eureka bundle contained chevron images used in `NavigationAccessoryView`. These chevrons can be replaced by the chevron SF Symbols in iOS 13 and up. For lower iOS versions they are fairly simple to draw in code using a few lines in CoreGraphics, which I've done also.

This should remove the blocker to adding Swift Package Manager support, so I've added a Package.swift file too.

The reimplemented chevrons look as follows (the left simulator uses the manually drawn ones, the right simulator uses SFSymbols):

Simulator screenshot showing chevrons


jimisaacs:
Looking forward to this.

Issue #552, opened 2016-07-11
Error: Duplicate tag when using custom object with PushRow


alexvazquez:
Hi,

I have been using custom objects with `AlertRow` without any problem.

Now, I tried to change `AlertRow` with `PushRow` and I get the following error:

`assertion failed: Duplicate tag DashPress.ServicePaymentRate: file /Users/demo/Documents/Development/Swift/Dashboard/dashboard main/Pods/Eureka/Source/Core/BaseRow.swift, line 146`

Here is the code:

```
<<< PushRow<ServicePaymentRate>() {
                $0.title = "Payment Rate"
                $0.selectorTitle = "Payment Rate"
                $0.options = self.servicePaymentRates
                
                $0.displayValueFor  = {
                    if let t = $0 {
                        return t.description 
                    }
                    return nil
                }
          
                }.onChange { row in
                    print(row.value)
                }
```
 Any clue on why Im getting this error using `PushRow` but it's working without error with `AlertRow`?

sunbohong:
Can you supply one demo?

alexvazquez:
Here is the demo where you can see that it works using AlertRow but you get the Duplicate tag error when using PushRow.

Thanks 

[EurekaIssueDemo.zip](https://github.com/xmartlabs/Eureka/files/359461/EurekaIssueDemo.zip)

mats-claassen:
This is strange. That error pops up when you are defining two rows with the same tag. Can you double check that you are not doing that?

alexvazquez:
I just give both rows difrerent tags and the issue still there:

 ```
<<< AlertRow<ServicePaymentRate>() {
                $0.tag  = "tag1"
                $0.title = "Payment Rate"
                $0.selectorTitle = "Select a Payment Rate"
                $0.options = self.servicePaymentRates
                
                $0.displayValueFor  = {
                    if let t = $0 {
                        return t.description //t.map({ countries[$0]! }).joinWithSeparator(", ")
                    }
                    return nil
                }
                //$0.value = 👦🏼
                }.onChange { row in
                    print(row.value)
                }
                .onPresent{ _, to in
                    //to.view.tintColor = .purpleColor()
                    //to.dispayValueFor = { return "Hi \($0)" }
                    
                }.cellUpdate { cell, row in
                    //row.options =  self.careTypes.map { $0.description }
            }
        
            
            
            <<< PushRow<ServicePaymentRate>() {
                $0.tag  = "tag2"
[...]

PR #2007, opened 2020-03-24
Append option index to the tags of the rows in SelectorViewController

rayx:
When using `PushRow` with custom type, if two instances of the custom type have same string representation, it would crash the App due to `"Duplicate tag"` error (see https://github.com/xmartlabs/Eureka/issues/552). As the author commented in that issue, the preferred method is to modify the custom type's `description` computed property to generate different string for different instance. However, that's not always feasible, given that the same string is also used as the text shown to user (meaning: one can't dump everything that helps to differentiate two instances in the string). In those (rare) cases, I think it's better to show those options with same text rather than crash the App. As far as I can tell, the only thing prevents that behavior is the tags generated in the code. The commit fixes that by making sure tags are unique.


Source/Rows/Controllers/SelectorViewController.swift
```diff
@@ -210,8 +210,8 @@ open class _SelectorViewController<Row: SelectableRowType, OptionsRow: OptionsPr
                 }
             }
         }
-        for option in options {
-            section <<< Row.init(String(describing: option)) { lrow in
+        for (index, option) in options.enumerated() {
+            section <<< Row.init(String(describing: option) + String(index)) { lrow in
```
mats-claassen: I think prefixing the tag with the index makes more sense. It is also still possible to have duplicate tags but that would be less common

rayx: Hi @mats-claassen How do you think about just using index as the tag? That should be unique and easy to read. Do you have any scenario in mind where it helps to include option value in the tag?

rayx: So I go ahead to use only option index as row tag value. Please look at the new diff. Let me know if you'd like to include option string in the tag. Thanks.

rayx: I'm sorry. I suppose only reviewers (not code change author) can click on "resolve conversion" button, right? I have clicked it again to "unresolve" it.

rayx:
@mats-claassen Could you please take a look at the new patch? The CI build failed. But from what I can tell, it didn't seem to have anything to do with my change (I tested the change in my own repo). Thanks.

mats-claassen:
It seems the unit tests failed. If you select the `Eureka` target scheme you can run them and see why they failed

rayx:
Thanks for the notification. I'm tied up with my project and won't be able to try it soon. Will look into it when I get a chance (perhaps in June).

Ashruss:
Tiny tuna

aPandeyisystango (approved):

aPandeyisystango (changes requested):
When I am using this piece of code and get rid of duplicate tag error.

rayx:
@aPandeyisystango I'm not sure what exact issue you ran into. I included the patch in one of my apps and it worked well. If you look at the code change, there is no reason why it could fail. Are you sure the issue you hit was the same issue the patch addressed? Anyway I haven't done iOS app development for a few months and probably won't do it any time soon, so I'm afraid I can't help on this.

Issue #2025, opened 2020-04-25
Example app crashes for all "Multiple Selection" / "Push Selection" row types.


intere:
### Relevant Context
- Xcode 11.4 (11E146)
- Eureka 5.2.0 `567d4e1 (HEAD, tag: 5.2.0) Remove bundle from podspec (#1981)`
    - This issue applies to everything 5.0.0 and up as well, probably predates the 5.0.0 baseline as well
- I can 100% reproduce this crash every time
    - I've found the culprit to be `[unowned self] in callback(self.cell, self)` and I'm going to submit a PR for this issue.

In trying to add a PushRow to my own app, I found a crash, so I decided to check out the example app.  I found that the example app is also crashing, so I dove a bit deeper.  I believe I know how to fix this and I'll work on submitting a PR.

### Example demonstration
![EurekaCrash](https://user-images.githubusercontent.com/2284832/80279994-54a18080-86be-11ea-95fc-fdf55bee111f.gif)

### Rows that reproduce this crash:
- `PushRow`
- `SectionedPushRow`
- `LazySectionedPushRow`
- `Custom Cell Push Row`
- `MultipleSelectorRow`
- `SectionedMultipleSelectorRow`
- `LazySectionedMultipleSelectorRow`

intere:
This issue is fixed by [This PR](https://github.com/xmartlabs/Eureka/pull/2026)

mats-claassen:
Have you tried Eureka 5.2.1?

intere:
I had the same issue on the master branch.  No, I did not try 5.2.1.

intere:
I didn't realize that 5.2.1 even existed.  Whoops.  🤦‍♂️

intere:
I'll close my PR, too.

Issue #2022, opened 2020-04-21
updateCell() not called on iOS 12.4.5


dneykov:
Hi,

today I discovered that on my older iPhone 6 using iOS 12.4.5 updateCell() is not called for some reason. I have PushRows in my form and after choosing an option and going back to the form the value is not shown. After scrolling down and going back the value is visible(seems that updateCell() is not called maybe?)

Eureka: **5.2.1**
Xcode: **11.4**

dneykov:
Calling `row.updateCell()` in `onChange` fixes the problem but that wasn't the case before.

mats-claassen:
I just tested a PushRow in the Examples and it works on Simulator iPhone 11, iOS 13.4. Does it work for you on iOS 13? Or on another iPhone?

dneykov:
Hi @mats-claassen,

yes it is working as expected on iOS 13. I can confirm that the problem appears in the examples as well on iOS 12. The problem is not only appearing on my iPhone but on iPad running iOS 12 as well.

As workaround I added on all my PushRows and MultipleSelectorRows `.onChange` callback.
I tried to fix it in my Pods so I can suggests a solution but did not manage

mats-claassen:
I am wondering if it has to do with this commit https://github.com/xmartlabs/Eureka/commit/2046ebf6a32e8d665942de5eb8a87b9056b462fa.

Could you test if using Eureka 5.1.0 in iOS 12 works? Or reverting the change in `Core.swift` of that commit to see if that fixes it for iOS 12. That would bring is closer to a fix/workaround

dneykov:
@mats-claassen reverting the change in `Core.swift` fix the problem. PushRows work as expected on iOS 12.

mats-claassen:
@dneykov I fixed it in #2029. As per your previous comment it should work for you

dneykov:
@mats-claassen as soon it's publish I will update, test and close this issue.

PR #2026, opened 2020-04-25
Issue #2025 - fixes crash on Push / Multi value rows.

intere:
### Background
As described in [Issue #2025](https://github.com/xmartlabs/Eureka/issues/2025), the example app (and apps that use the Push or MultiSelection row types) crash when you tap on those row types.

This PR serves to fix this issue by replacing the use of `[unowned self]` with `[weak self]` in the `RowType.swift` file for the callback blocks.  I can't comment on the exact reason this went from "working" to "crashing", but I know that this PR fixes that crash.

### Breaking Changes
I don't know how this worked before, but the Cells do not refresh automatically when you selected updated values in the example app, but if you scroll them out of view and scroll back to them, they do reflect the updated selection.  I don't know if this is how it behaved before or not, so this is potentially a breaking change, though it's also an improvement in that the app doesn't crash.

***Update*** - The "show updated state" after selecting a push or multi selection works without any issue in the app that I'm building, I'm not sure why it doesn't work properly in the example app.

### Metadata
- Issue(s): https://github.com/xmartlabs/Eureka/issues/2025

### Screenshots
Demonstration of the app not crashing after the fix:

EurekaCrashFix

### DIFF
--- a/Source/Core/RowType.swift
+++ b/Source/Core/RowType.swift
@@ -202,7 +202,12 @@ extension RowType where Self: BaseRow {
      */
     @discardableResult
     public func onChange(_ callback: @escaping (Self) -> Void) -> Self {
-        callbackOnChange = { [weak self] in callback(self!) }
+        callbackOnChange = { [weak self] in
+            guard let self = self else {
+                return
+            }
+            callback(self)
+        }
         return self
     }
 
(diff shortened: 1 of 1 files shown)


mats-claassen:
Is that screenshot on iOS 12? I think it is related to #2022

intere:
It turns out, this was already fixed in 5.2.1, closing this PR.

PR #2029, opened 2020-04-27
Reload rows on viewWillAppear for iOS 12 and below.

### LINKED ISSUE #2022: updateCell() not called on iOS 12.4.5

mats-claassen:
Fixes #2022


chitvan832:
@mats-claassen I don't see these changes in latest version of Eureka.

mats-claassen:
It hasn't been released yet. You can use the master branch until the next release

PR #2046, opened 2020-06-03
Start editing field rows at the end when tapping blank space in row

### LINKED ISSUE #2044: Propose to make cursor starting from the right end of current text for TextRow
Just noticed when tap blank area of a TextRow to edit, the cursor will start blinking at the left end of already-entered text. That can cause trouble to users, considering it's not easy to move cursor on a mobile device by dragging.

I think this is due to the nature of UITextField. It's natural when the text is left-aligned, as most of Apple apps use it that way. But Eureka uses right-aligned text field.

One solution is to make cursor appear at the right end by default when clicking blank area. If click on text, then follow the original behavior.

mats-claassen:
Fixes #2044

### DIFF
--- a/Source/Rows/Common/FieldRow.swift
+++ b/Source/Rows/Common/FieldRow.swift
@@ -206,9 +206,11 @@ open class _FieldCell<T> : Cell<T>, UITextFieldDelegate, TextFieldCell where T: 
         }
         textField.addTarget(self, action: #selector(_FieldCell.textFieldDidChange(_:)), for: .editingChanged)
 
-        // Make sure the title takes over most of the empty space so that the text field starts editing at the back.
-        let priority = UILayoutPriority(rawValue: titleLabel!.contentHuggingPriority(for: .horizontal).rawValue + 1)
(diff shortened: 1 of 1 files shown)


mtnbarreto:
```
- (void) textViewDidBeginEditing:(UITextView *)textView {
    textView.selectedRange = NSMakeRange(0, 0);
} 
```

may also work

Source/Rows/Common/FieldRow.swift
```diff
@@ -206,6 +206,9 @@ open class _FieldCell<T> : Cell<T>, UITextFieldDelegate, TextFieldCell where T:
         }
         textField.addTarget(self, action: #selector(_FieldCell.textFieldDidChange(_:)), for: .editingChanged)
 
+        // Make sure the title takes over most of the empty space so that the text field starts editing at the back.
+        let priority = UILayoutPriority(rawValue: titleLabel!.contentHuggingPriority(for: .horizontal).rawValue + 1)
```
mtnbarreto, line 210: avoid `!`

PR #2061, opened 2020-07-23
Workaround for crash in Swift 5.3 beta 2/3

mattgallagher:
This PR provides a workaround to the crash affecting Swift 5.3 betas 2 and 3. It replaces a `Self: Section` constraint with two alternate constraints that allow the code to achieve the same result.

Working around bugs in beta compilers isn't always the best approach so I understand if this doesn't get merged. It's mostly an FYI for anyone looking to get around the crash in the meantime.

### DIFF
--- a/Source/Core/SelectableSection.swift
+++ b/Source/Core/SelectableSection.swift
@@ -62,7 +62,7 @@ public protocol SelectableSectionType: Collection {
     func selectedRows() -> [SelectableRow]
 }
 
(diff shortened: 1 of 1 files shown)


mats-claassen:
Thanks @mattgallagher for the workaround. I have merged in into a xcode12 branch. let's see if this issue is fixed before the final Xcode version is released. If not, we'll merge this into master

PR #2067, opened 2020-08-25
Fix DateInlineRow, TimeInlineRow and DateTimeInlineRow style/rend…

kamerc:
Fix DateInlineRow, TimeInlineRow and DateTimeInlineRow style/rendering issues for iOS14. See screenshots from the Example app for details. The default picker view is .inline to match the iOS native apps. If a user is on iOS 13 or lower, they will still see the default style of wheels

### DateInlineRow (before)
dateBefore
### DateInlineRow (after)
dateAfter

### TimeInlineRow (before)
timeBefore
### TimeInlineRow (after)
timeAfter

### DateTimeInlineRow (before)
dateTimeBefore
### DateTimeInlineRow (after)
dateTimeAfter

### DIFF
--- a/Source/Rows/DateInlineRow.swift
+++ b/Source/Rows/DateInlineRow.swift
@@ -118,7 +118,7 @@ open class _CountDownInlineRow: _DateInlineFieldRow {
 public final class DateInlineRow_<T>: _DateInlineRow, RowType, InlineRowType {
     required public init(tag: String?) {
         super.init(tag: tag)
-        onExpandInlineRow { cell, row, inlineRow in
+        onExpandInlineRow { cell, row, _ in
             let color = cell.detailTextLabel?.textColor
             row.onCollapseInlineRow { cell, _, _ in
                 cell.detailTextLabel?.textColor = color
@@ -141,7 +141,7 @@ public typealias DateInlineRow = DateInlineRow_<Date>
 public final class DateTimeInlineRow_<T>: _DateTimeInlineRow, RowType, InlineRowType {
     required public init(tag: String?) {
         super.init(tag: tag)
-        onExpandInlineRow { cell, row, inlineRow in
+        onExpandInlineRow { cell, row, _ in
             let color = cell.detailTextLabel?.textColor
             row.onCollapseInlineRow { cell, _, _ in
                 cell.detailTextLabel?.textColor = color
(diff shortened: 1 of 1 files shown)


funkenstrahlen:
Thank you!

Source/Rows/DateInlineRow.swift
```diff
@@ -105,7 +105,16 @@ open class _CountDownInlineRow: _DateInlineFieldRow {
 public final class DateInlineRow_<T>: _DateInlineRow, RowType, InlineRowType {
     required public init(tag: String?) {
         super.init(tag: tag)
-        onExpandInlineRow { cell, row, _ in
+        onExpandInlineRow { cell, row, inlineRow in
+            inlineRow.cellUpdate() { cell, row in
+                cell.datePicker.datePickerMode = .date
```
mats-claassen, line 110: I think setting datePickerMode is not needed here, right?

mats-claassen:
I agree with the change but I think `onExpandInlineRow` isn't the best place for this as anyone who overrides this has to copy this code as well to maintain the functionality. From what I saw, if we remove the line setting the `datePickerMode` it is the same code for each date row except CountDownRow which does not support `.inline` mode. What about adding this change to `configureInlineRow` in `DatePickerRowProtocol`? You will have to check whether `datePickerMode` is countdown to avoid applying this change to CountDown rows

Source/Rows/DateInlineRow.swift
kamerc: From the documentation: default is UIDatePickerModeDateAndTime. So this line needs to stay for DateInlineRow and also the line that sets it to .time on the TimeInlineRow. But it looks like the line for DateTimeInlineRow that sets it to .dateAndTime can be removed.

kamerc:
@mats-claassen configureInlineRow takes DatePickerRowProtocol rather than DatePickerRow, which I need to be able to access the cell.datePicker. What if I add a new method to call from setupInlineRow that passes DatePickerRow?

mats-claassen:
Sounds good. Then just don't call it in CountDown row

kamerc:
@mats-claassen I have pushed an update

Source/Rows/DateInlineRow.swift
mats-claassen: I mean `datePickerMode` is set in `DatePickerCell` (line 54). It is not necessary to re-set it here

PR #2077, opened 2020-09-17
datePickerStyle for DateFieldRow and DatePickerCell

tscheepers:
Analogous to #2067 but for DateFieldRow and DatePickerCell.

For DateFieldRow the accessory view did use the wrong date picker, this should always be the wheel view. For DatePickerCell the compact view should be ignored similar to #2067.


mats-claassen:
Please have a look at #2078 and how this change breaks both building for Mac Catalyst and Xcode prior to 11.4

tscheepers:
@mats-claassen I've added the conditionals similar to #2078 :)

Source/Rows/DatePickerRow.swift
```diff
@@ -53,6 +53,18 @@ open class DatePickerCell: Cell<Date>, CellType {
         height = { UITableView.automaticDimension }
         datePicker.datePickerMode = datePickerMode()
         datePicker.addTarget(self, action: #selector(DatePickerCell.datePickerValueChanged(_:)), for: .valueChanged)
+
+        #if swift(>=5.2)
+            if #available(iOS 14.0, *) {
+                #if swift(>=5.3) && !(os(OSX) || (os(iOS) && targetEnvironment(macCatalyst)))
+                    cell.datePicker.preferredDatePickerStyle = .inline
```
mats-claassen, line 60:
```suggestion
                    datePicker.preferredDatePickerStyle = .inline
```

Source/Rows/DatePickerRow.swift
```diff
@@ -53,6 +53,18 @@ open class DatePickerCell: Cell<Date>, CellType {
...
+                #else
+                    cell.datePicker.preferredDatePickerStyle = .wheels
```
mats-claassen, line 62:
```suggestion
                    datePicker.preferredDatePickerStyle = .wheels
```

mats-claassen (changes requested):

mats-claassen (approved):

Source/Rows/DatePickerRow.swift
```diff
@@ -53,6 +53,18 @@ open class DatePickerCell: Cell<Date>, CellType {
         height = { UITableView.automaticDimension }
         datePicker.datePickerMode = datePickerMode()
         datePicker.addTarget(self, action: #selector(DatePickerCell.datePickerValueChanged(_:)), for: .valueChanged)
+
+        #if swift(>=5.2)
+            if #available(iOS 14.0, *) {
+                #if swift(>=5.3) && !(os(OSX) || (os(iOS) && targetEnvironment(macCatalyst)))
+                    datePicker.preferredDatePickerStyle = .inline
```
funkenstrahlen, line 60: @mats-claassen Why using `.inline` here? Should this not be exclusive for `DateInlineRow`?

I explicitly try to avoid the new iOS 14 DatePicker with inline style because the time selection is covered by the keyboard when being edited. Maybe a contentInset issue.

PR #2105, opened 2020-10-24
Add replace method for all sections

### LINKED ISSUE #2003: Reloading sections the correct way
Hey there.

We use Eureka in an app which uses MVVM, so the setup of the data source (section/ rows) is set up in the view models. The said data source is not static, e.g. sections might be replaced, once the value of a picker changes.
Once the data source is being updated, we used to use `Form.replaceSubrange` in the respective view controller to just replace all existing sections with the new ones, like: 

```
form.replaceSubrange(0..<self.form.allSections.count, with: sections)
```

This would be fine for us, even if it doesn't perform any diffing or something. I know that we would have to implement diffing our own.
The issue here is, `Form`s implementation of `replaceSubrange(_:with:)` doesn't remove any sections which are currently hidden ([see this line](https://github.com/xmartlabs/Eureka/blob/master/Source/Core/Form.swift#L218); because `sections` apparently doesn't include the hidden sections).

That means: once I call `replaceSubrange` while some of the sections are hidden, the hidden ones do not get removed. As a result, there will be more sections in the form, after enabling the switch which is responsible for hiding said sections.

So to the actual questions:
Is is intended to use `sections` in the linked line, which does ignore the hidden sections, or would use `_allSections` make more sense?
Is there a better (or at least another) way to replace all (including hidden) sections which I have overlooked?

Thanks for your time.

forceunwrap:
This PR adds a new method `Form.replaceSubrangeInAllSections(_:with:)` which replaces a subrange of *all* sections (including hidden ones) with a given set of sections.
In addition, it *replaces* existing visible sections within the subrange by the visible sections in the new element collection.

@mats-claassen Please let me know if you have any concerns.

Resolves #2003

### DIFF
--- a/Source/Core/Form.swift
+++ b/Source/Core/Form.swift
@@ -238,23 +238,37 @@ extension Form : RangeReplaceableCollection {
         _ subRange: Range<Int>,
         with newElements: C
     ) where C.Iterator.Element == Section {
-        let newVisibleElements = newElements.filter { !$0.isHidden }
+        // Keep a reference to all removed sections in order to create a subrange for the visible section replacement
+        var removedSections = [Section]()
 
         // Remove subrange in all sections
         for i in subRange.reversed() where kvoWrapper._allSections.count > i {
             let removed = kvoWrapper._allSections.remove(at: i)
             removed.willBeRemovedFromForm()
+            removedSections.insert(removed, at: 0)
         }
 
-        // Replace subrange in visible sections using a new subrange
-        let visibleLength = Swift.min(subRange.upperBound, kvoWrapper.sections.count) - subRange.lowerBound
-        let visibleSubrange = NSRange(location: subRange.lowerBound, length: visibleLength)
-        kvoWrapper.sections.replaceObjects(in: visibleSubrange, withObjectsFrom: newVisibleElements)
+        let removedVisibleSections = removedSections.filter { !$0.isHidden }
 
-        // Insert all new elements into all sections
+        // Get indices of first and last section which is not hidden...
+        let firstVisible = kvoWrapper.sections.contains(removedVisibleSections.first) ? kvoWrapper.sections.index(of: removedVisibleSections.first) : 0
+        let lastVisible = kvoWrapper.sections.contains(removedVisibleSections.last) ? kvoWrapper.sections.index(of: removedVisibleSections.last) : 0
+
+        // ...and create a range from those indices
+        let range = firstVisible...lastVisible
+
+        let newVisibleSections = newElements.filter { !$0.isHidden }
+
+        // Replace visible form sections
+        kvoWrapper.sections.replaceObjects(
+            // Create a new NSRange using location and length because interop between `ClosedRange` and `NSRange` is bad
+            in: NSRange(location: range.lowerBound, length: Swift.min(kvoWrapper.sections.count, range.count)),
+            withObjectsFrom: newVisibleSections
+        )
+
         kvoWrapper._allSections.insert(contentsOf: newElements, at: indexForInsertion(at: subRange.lowerBound))
 
-        for section in newVisibleElements {
+        for section in newVisibleSections {
             section.wasAddedTo(form: self)
         }
     }
--- a/Source/Core/Form.swift
+++ b/Source/Core/Form.swift
@@ -213,18 +213,48 @@ extension Form : RangeReplaceableCollection {
         }
     }
 
-    public func replaceSubrange<C: Collection>(_ subRange: Range<Int>, with newElements: C) where C.Iterator.Element == Section {
+    public func replaceSubrange<C: Collection>(
+        _ subRange: Range<Int>,
+        with newElements: C
+    ) where C.Iterator.Element == Section {
         for i in subRange.lowerBound..<subRange.upperBound {
             if let section = kvoWrapper.sections.object(at: i) as? Section {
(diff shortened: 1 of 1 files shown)


Source/Core/Form.swift
```diff
@@ -213,22 +213,66 @@ extension Form : RangeReplaceableCollection {
...
+        kvoWrapper.sections.replaceObjects(
+            // Create a new NSRange using location and length because interop between `ClosedRange` and `NSRange` is bad
+            in: NSRange(location: range.lowerBound, length: Swift.min(kvoWrapper.sections.count, range.count)),
+            withObjectsFrom: newVisibleSections
+        )
+
+        kvoWrapper._allSections.insert(contentsOf: newElements, at: indexForInsertion(at: subRange.lowerBound))
+
+        for section in newVisibleSections {
+            section.wasAddedTo(form: self)
```
mats-claassen, line 272: Actually this should be called on all sections. It will also evaluate which sections are hidden and remove them from kvowrapper.sections so you can safely insert all sections into both collections.

Source/Core/Form.swift
```diff
@@ -213,22 +213,66 @@ extension Form : RangeReplaceableCollection {
...
+        for i in subRange.reversed() where kvoWrapper._allSections.count > i {
+            let removed = kvoWrapper._allSections.remove(at: i)
+            removed.willBeRemovedFromForm()
+            removedSections.insert(removed, at: 0)
+        }
+
+        let removedVisibleSections = removedSections.filter { !$0.isHidden }
+
+        // Get indices of first and last section which is not hidden...
+        let firstVisible = kvoWrapper.sections.contains(removedVisibleSections.first) ? kvoWrapper.sections.index(of: removedVisibleSections.first) : 0
```
mats-claassen, line 254: I know these sections should be in the kvoWrapper.sections array but the fallback of 0 does not look good in case something does not behave as expected. You could create a range like (X, 0) which would be invalid or (0, X) which  would remove more sections than wanted. What about an if let?

Source/Core/Form.swift
```diff
@@ -213,22 +213,66 @@ extension Form : RangeReplaceableCollection {
...
+
+        // ...and create a range from those indices
+        let range = firstVisible...lastVisible
+
+        let newVisibleSections = newElements.filter { !$0.isHidden }
+
+        // Replace visible form sections
+        kvoWrapper.sections.replaceObjects(
+            // Create a new NSRange using location and length because interop between `ClosedRange` and `NSRange` is bad
+            in: NSRange(location: range.lowerBound, length: Swift.min(kvoWrapper.sections.count, range.count)),
```
mats-claassen, line 265: length should be min of range.count and `kvoWrapper.sections.count` - `range.lowerBound`, right?

Source/Core/Form.swift
```diff
@@ -213,22 +213,66 @@ extension Form : RangeReplaceableCollection {
...
+        let removedVisibleSections = removedSections.filter { !$0.isHidden }
+
+        // Get indices of first and last section which is not hidden...
+        let firstVisible = kvoWrapper.sections.contains(removedVisibleSections.first) ? kvoWrapper.sections.index(of: removedVisibleSections.first) : 0
+        let lastVisible = kvoWrapper.sections.contains(removedVisibleSections.last) ? kvoWrapper.sections.index(of: removedVisibleSections.last) : 0
+
+        // ...and create a range from those indices
+        let range = firstVisible...lastVisible
+
+        let newVisibleSections = newElements.filter { !$0.isHidden }
```
mats-claassen, line 260: You can insert all sections into `kvoWrapper.sections` as `willAddTo(form:)` will call `evaluateHidden` on them. `isHidden` will most likely be false for all of them anyway.

Source/Core/Form.swift
forceunwrap: ✅

Source/Core/Form.swift
forceunwrap: Thinking about this and how to solve it took me way longer than I thought it would. 😁
The range which would have been created should have never been invalid (in terms of crash-safety), because `last >= first` should always be true in case of indices. 
Still the fallback of `0` doesn't make sense,  as you have already mentioned. Some potential problems would have been:
- Replacing only hidden sections at the end of the array with a set of visible sections
- Replacing only hidden sections *in between* two visible sections

I first tried to come up with a solution which handles all of the problematic scenarios. But then I suddenly had a brainwave, where I remembered you told me that all hidden sections are going to be removed anyway.

So the most easy solution is probably just to replace all visible sections by `_allSections` after processing it. This works for us. I hope this makes sense.

mats-claassen (approved):

mats-claassen:
The changes look good. I think it would be good to add one or two unit tests for this function. For example something like this:

```swift
func testReplaceAllSection() {
        let form = Form() +++ Section("section1") {
            $0.hidden = true
        }
            +++ Section("section2")
            +++ Section("section3")

        form.replaceSubrangeInAllSections(Range<Int>(uncheckedBounds: (lower: 0, upper: 2)), with: [Section("section0") { $0.hidden = true }])

        XCTAssertEqual(form.allSections.count, 2)
        XCTAssertEqual(form.count, 1)
        XCTAssertEqual(form[0].header?.title, "section3")
        XCTAssertEqual(form.allSections[0].header?.title, "section0")
        XCTAssertEqual(form.allSections[1].header?.title, "section3")
    }
```

PR #2129, opened 2020-12-21
Modify hugging priority for PasswordRow

### LINKED ISSUE #2125: RE: PasswordRow TextField Display ommited by three-pointed reader.
I previously reported an issue where the password field was omitted with "..." in 3.5.0.
This time I checked out 3.5.2 and rechecked to see that the issue wasn't resolved.

Please see the capture.
If the password text becomes long, it will be omitted with "...".
After all, commenting out what I pointed out earlier solved the problem.

`/* if let titleLabel = titleLabel { ... } */`

Hmm. Type "ijklijkli" and it will be omitted.
"ijkl" can be entered up to 2 times, but when "i" is added, "..." is displayed at the end.
The image below is all entered as "ijklijkli".

Please check again.
Thank you.

EUREKA_5 3 2_IMG_0456

_Originally posted by @vertuxx in https://github.com/xmartlabs/Eureka/issues/2079#issuecomment-745196343_

mats-claassen:
Fixes #2125

### DIFF
--- a/Source/Rows/FieldsRow.swift
+++ b/Source/Rows/FieldsRow.swift
@@ -138,7 +138,9 @@ open class PasswordCell: _FieldCell<String>, CellType {
         textField.autocapitalizationType = .none
         textField.keyboardType = .asciiCapable
         textField.isSecureTextEntry = true
-        textField.adjustsFontSizeToFitWidth = true
+        if let textLabel = textLabel {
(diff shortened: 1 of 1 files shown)


vertuxx (approved):
Hello @mats-claassen
I've seen the modified code in FieldsRow.swift.
It seems that there is no problem running Example on iPhone.
It looks okay.

Thank you.

PR #2148, opened 2021-03-27
Fix #1952: Use SafeArea Autolayout Constraints when FormViewController Is Presented Modally

### LINKED ISSUE #1952: ScrollView contentOffset not adjusted correctly when form is displayed as page sheet and keyboard is displayed

philipbel:
`UITableView.scrollRectToVisible(, animated:)` and `UITableView.scrollToRow()` don't work properly in `keyboardWillShow()` due to, what seems to be, a bug in iOS when autoresizing mask is used for a `UIViewController` that is presented modally as `.formSheet/.pageSheet` within a `UINavigationController`.


Source/Core/Core.swift
```diff
@@ -458,12 +458,36 @@ open class FormViewController: UIViewController, FormViewControllerProtocol, For
...
+         */
+        var useAutoresizingMask = true
+        if #available(iOS 13, *) {
+            if presentingViewController != nil && navigationController != nil {
+                useAutoresizingMask = false
+
+                // Need to set the background color as otherwise black shows underneath the translucent navigation bar.
+                view.backgroundColor = .systemBackground
+                tableView.translatesAutoresizingMaskIntoConstraints = false
+                NSLayoutConstraint.activate([
```
mats-claassen, line 480: Hi, thanks for contributing. First comment is that a FormViewController can be created from a Storyboard where a UITableView is added with constraints (which might be equal to these but might as well not). The case where no Storyboard is used and the view controller is instantiated from code is handled in the `if tableView == nil {` above. As I understand, this issue only happens in that case, so any fix/workaround we apply should only apply to that case.

philipbel: Thanks for the comment, Mathias.

I have updated the PR. Moved the constraints to a new function and I believe I am handling the Storyboard/non-Storyboard case now.

Is this more in-line with what you have in mind?

PR #2150, opened 2021-03-29
Fix Chinese input issue

azone:
do nothing if the text field has markedTextRange


mats-claassen:
Please explain the issue, ideally with code to reproduce and state versions of Eureka, Xcode, iOS

azone:
> If there is no marked text, the value of the property is nil. Marked text is provisionally inserted text that requires user confirmation; it occurs in multistage text input. The current selection, which can be a caret or an extended range, always occurs within the marked text. (https://developer.apple.com/documentation/uikit/uitextinput/1614489-markedtextrange)

If you want to reproduce it, just run the demo, and do the following steps:
1. input any Chinese characters and confirm the marked text
2. input Chinese characters again

After those two steps, you'll found that you can't input anymore, or very weird.

mats-claassen (approved):

PR #2168, opened 2021-07-14
UIDatePicker Locale and TimeZone Options

hslightnin:
Added option to set locale to force 12/24 hour time input and option to set timezone of the picker.
This can be used for example to input times using a standard location i.e Zulu/UTC


mats-claassen:
Hi @hslightnin thanks for contributing but this feature is already possible like this:

```swift
<<< DateRow() {
    $0.value = Date();
    $0.title = "DateRow"
}.cellSetup({ (cell, row) in
    cell.datePicker.timeZone = TimeZone(identifier: "UTC")
    cell.datePicker.locale = Locale(identifier: "en_GB")
})
```

So I think it is not necessary to add any new variables.

hslightnin:
Awesome thanks!

PR #2205, opened 2022-01-17
Fix crash when user taps tab key on external keyboard

### LINKED ISSUE #2181: Crash on ipadOS/iOS 15 when pressing Tab key using hardware keyboard
Reproducible crash on ipadOS/iOS 15 when pressing Tab key multiple times using hardware keyboard (bluetooth, or usb-c)

Environment: iPad Pro 11, iPad OS 15.0.0, Eureka 5.3.3

```
*** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'Received a call to becomeFirstResponder while processing a focus update, which is not supported.'
terminating with uncaught exception of type NSException
```

in Core.swift:

image

Could be related to the new iOS 15 focus system, or something similar...

mats-claassen:
Fixes #2181

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,9 +11,12 @@ jobs:
     runs-on: macOS-latest
     strategy:
       matrix:
-        destination: ['platform=iOS Simulator,OS=14.4,name=iPhone 11']
+        destination: ['platform=iOS Simulator,OS=15.2,name=iPhone 11']
 
     steps:
     - uses: actions/checkout@v2
+    - uses: maxim-lobanov/setup-xcode@v1
+      with:
+        xcode-version: '13.2.1'
     - name: Build and test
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -11,9 +11,12 @@ jobs:
     runs-on: macOS-latest
     strategy:
       matrix:
-        destination: ['platform=iOS Simulator,OS=14.4,name=iPhone 11']
+        destination: ['platform=iOS Simulator,OS=15.2,name=iPhone 11']
```
trackvia-ios, line 14: I assume this updates your framework to be built against iOS 15.2 and Xcode version 13.2.1?

mats-claassen: Yes, but just for the tests

PR #2225, opened 2022-07-13
[Bug] Clear Form once view Did Disappear to prevent leak

CRoig:
Particular sections are retained by KVO Wrapper. In order to avoid memory leaks they should be removed explicitly before attempting deallocating VC.


mats-claassen:
Hi @CRoig could you please explain more about the leak and/or link an issue if there is one that does this?

freelancerios (approved):

PR #2242, opened 2023-02-18
@FormBuilder

RomanPodymov:
Hello.
Thank you for Eureka.
Add `@FormBuilder`.


README.md
```diff
@@ -228,6 +228,24 @@ form +++ {
 }
 ```
 
+#### @FormBuilder
+```swift
+@FormBuilder
+var form: Form {
+    nil as Section?
```
mats-claassen, line 235: Why this?

README.md
```diff
@@ -228,6 +228,24 @@ form +++ {
...
+    Section("Section A") { section in
+        section.tag = "Section_A"
+    }
+    if true {
+        Section("Section B") { section in
+            section.tag = "Section_B"
+        }
+    }
+    nil as TextRow?
```
mats-claassen, line 244: What does this do?

RomanPodymov: Hello @mats-claassen 
Fixed.

mats-claassen: Hi, so my question is why do we need to put a nil there?

RomanPodymov: What about `if something { nil }` ?

RomanPodymov: Hello @mats-claassen 
OK, I can remove
```
public static func buildOptional(_ components: [RowsProvider]?) -> [BaseRow] {
    components?.flatMap { $0.rows } ?? []
}
```
and
```
public static func buildOptional(_ components: [SectionsProvider]?) -> [Section] {
    components?.flatMap { $0.sections } ?? []
}
```

RomanPodymov: Or should I leave them and remove `if something { nil }` from the examples?

mats-claassen: I think buildOptional is correct. I just don't think that it is an intuititve example to add `nil` to a form. You could add an optional variable in the example which could be nil but there is no case when you would explicitly put `nil` there, right?

RomanPodymov: Hello @mats-claassen 
Exactly. So, should we remove it from the examples?

RomanPodymov: Or you can add your suggestions to my branch.

RomanPodymov: Hello @mats-claassen 
Can you please tell me what else should I fix in this pull request?

mats-claassen: Hey @RomanPodymov I think if you remove the `if something { nil}` from the examples then this is fine to be merged

RomanPodymov: Hello @mats-claassen 
Done.

RomanPodymov:
Hello @mats-claassen 
What other changes should I make in this pull request?

PR #2245, opened 2023-05-11
fix bug where tapping TextAreaRow always puts the cursor at the end

littlebobert:
on iOS 16, when I tap on a TextAreaRow’s content to start editing, the cursor always goes to the end.

I can reproduce this in Eureka’s sample app. here’s a video:

[opting out of TextKit 2](https://developer.apple.com/videos/play/wwdc2022-10090/?time=643) inside _TextAreaCell’s UITextView resolves the issue.


michaelpahler:
ahh having the same issue

PR #2246, opened 2023-05-23
Add displayValueForFirstRow and displayValueForSecondRow closure in DoublePickerInlineRow

lilidotshi:
Currently there's no way to configure the way that an DoublePickerInlineRow displays its values in the picker, which is avaliable in DoublePickerRow and DoublePickerInputRow. This pull request addresses that missing feature.

After:
Screenshot 2023-05-23 at 3 23 49 PM


lilidotshi:
@mats-claassen any thoughts on this PR? Should be pretty simple, just bringing the inline picker up to parity with the others. Our project is blocked by this otherwise, unless we write our own cell.

mats-claassen (approved):

PR #2248, opened 2023-08-25
Improve URL pattern matching

zero0cool0:
The current URL regex has a number of issues, for example, it doesn't match `http://localhost` or `https://example.com:8080`.

The proposed expression is taken from [here](https://stackoverflow.com/a/22648406)

It passes all the validation tests including the newly added ones.


zero0cool0:
@mats-claassen Any chance this can get merged into master?

mats-claassen (approved):

